PR #18, opened 2015-02-08
Correctly serialize ImmutableRoaringBitmap 

bpot:
It was writing the entire backing array even if the ByteBuffer only covered part of the array.

### DIFF
--- a/src/test/java/org/roaringbitmap/buffer/TestImmutableRoaringBitmap.java
+++ b/src/test/java/org/roaringbitmap/buffer/TestImmutableRoaringBitmap.java
@@ -22,6 +22,20 @@ import java.util.Random;
 public class TestImmutableRoaringBitmap {
     
     @Test
+    public void testSerializeImmutable() throws IOException {
+        MutableRoaringBitmap mr = new MutableRoaringBitmap();
+        mr.add(5);
+        ByteBuffer buffer = serializeRoaring(mr);
+        
+        buffer.rewind();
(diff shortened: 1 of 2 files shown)


src/test/java/org/roaringbitmap/buffer/TestImmutableRoaringBitmap.java
```diff
@@ -82,18 +96,19 @@ public void testContains() throws IOException {
...
 	            public void write(int b) {mBB.put((byte) b);}
-	            public void write(byte[] b) {}            
-	            public void write(byte[] b, int off, int l) {}
+	            public void write(byte[] b) {}
+	            public void write(byte[] b, int off, int l) { mBB.put(b,off,l); }
 	        }.init(outbb));
 			mrb.serialize(dos);
 			dos.close();
-			
+
```
bpot: I'd like to cleanup some of the indentation in this method but wasn't sure if I should do that at the same time as functionality changes.

lemire: Please go ahead. But let me have a look at your bug fix first. If I can, I will merge it quickly.

lemire:
I am merging this, but I think you may have found a larger issue because I think it is assumed throughout that the offset is zero.

lemire:
I just added small fix to make sure that we check arrayOffset throughout in the code. Thanks for catching this.

PR #19, opened 2015-03-10
Flyweight primitive Iterator over a RoaringBitmap.

bobymicroby:
This Iterator will let you go green and make your garbage collector suffer less.

 *  Benchmark                                                      Mode    Samples    Score   Score error  Units
+ b.c.b.t.b.roaringBitmap.RoaringIteratorBenchmark.testFlyweight    thrpt      200  17347.182      102.962  ops/s
+ b.c.b.t.b.roaringBitmap.RoaringIteratorBenchmark.testStandard     thrpt      200  10302.179       53.562  ops/s

### DIFF
--- a/src/main/java/org/roaringbitmap/IntIteratorFlyweight.java
+++ b/src/main/java/org/roaringbitmap/IntIteratorFlyweight.java
@@ -1,0 +1,256 @@
+package org.roaringbitmap;
+
+/**
+ * This Iterator will let you go green and make your garbage collector suffer less.
+ *
+ * You can create one reusable instance of this class and then {@link #wrap(RoaringBitmap)}
+ *
+ * I am aware of the code duplication and the VeryLongNamingOfTheIterators but in order
+ * to fix this it will take to make a lot of changes to the other Iterators in this project as well.
+ *
+ *
+ *
+ *
+ *
+ *
+ * Here is the attached benchmark
+ * public class RoaringIteratorBenchmark {
+
+
+
+ public int  testStandard(RoaringState state) {
+
+ IntIterator intIterator = state.bitmap.getIntIterator();
+ int result =0;
+ while(intIterator.hasNext()) {
+ result = intIterator.next();
+
+ }
+ return result;
+
+ }
+
+
+ @Benchmark
+ public int  testFlyweight(RoaringState state) {
+
+ IntIteratorFlyweight intIterator = state.intIterator;
+
+ intIterator.wrap(state.bitmap);
+
+ int result =0;
+ while(intIterator.hasNext()) {
+ result = intIterator.next();
+
+ }
+ return result;
+
+ }
+
+
+
+ @State(Scope.Benchmark)
+ public static class RoaringState {
+
+ final Random source = new Random(0xcb000a2b9b5bdfb6l);
+ final int[] data = takeSortedAndDistinct(source, 10000);
+
+ final RoaringBitmap bitmap = RoaringBitmap.bitmapOf(data);
+
+ final IntIteratorFlyweight intIterator = new IntIteratorFlyweight();
+
+
+
+ private static int[] takeSortedAndDistinct(Random source, int count) {
+ LinkedHashSet<Integer> ints = new LinkedHashSet<Integer>(count);
+ for (int size = 0; size < count; size++) {
+ int next;
+ do {
+ next = Math.abs(source.nextInt());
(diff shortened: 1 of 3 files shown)


lemire:
Thank you! I will have a look soon!

lemire:
This looks good to me. I merged it since it is non-intrusive.

I will then take a closer look.

Ideally, I would want to add your name as an author... If you could share your name, that would be great.

Thanks.

bobymicroby:
I am evaluating the bitmaps for a real-time latency sensitive application and for now I must say I am pretty satisfied with the results. I am doing everything possible to squeeze few cycles here and there.
My main goal is to minimize the object allocation to a minimum.

Locally I have changed/benchmarked/tested changing the sorting in FastAggregation 

from this
 ```java    
Arrays.sort(array, new Comparator<RoaringBitmap>() {
            @Override
            public int compare(RoaringBitmap a, RoaringBitmap b) {
                return a.getSizeInBytes() - b.getSizeInBytes();
            }
        }); 

```
to this
```java

 private void insertionSort(RoaringBitmap[] a, int left, int right) {

            for (int i = left, j = i; i < right; j = ++i) {
               RoaringBitmap ai = a[i + 1];
               while (ai.getCardinality() < a[j].getCardinality()) {
                  a[j + 1] = a[j];
                  if (j-- == left) {
                     break;
                  }
               }
               a[j + 1] = ai;
            }

   }
```

I am using insertion sort because I never merge more than 50 mapsThis change gives another 50,000+ add ops/second. I have one question in regard to this change:

I have benchmarked getCardinality() to give better throughput than getSizeInBytes(). Do you think it is reasonable to use getCardinality for the sorting operation ? 

P.S My name is Borislav Ivanov.

lemire:
I have permanently added your name to the list of authors.

This being said, I am not sure that we should keep your new classes. Can you check my work? I have written a benchmark of your code and I get, at best, inclusive results, and at worse, I get negative results. I do not use the same benchmark platform as you used, but this should not matter.

It is true that you reduce the number GC calls, but I have no evidence here that GC calls are a significant burden. And for some unknown reason, your code appears to be overall slower.

You can reproduce my results from a fresh pull...

      $ mvn -Dtest=TestBenchmarkIterator test
      Running org.roaringbitmap.TestBenchmarkIterator
      TestBenchmarkIterator.testFlyweight: [measured 50000 out of 50003 rounds, threads: 1 (sequential)]
       round: 0.00 [+- 0.00], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 0, GC.time: 0.00, time.total: 0.68, time.warmup: 0.00, time.bench: 0.68
      TestBenchmarkIterator.testStandard: [measured 50000 out of 50003 rounds, threads: 1 (sequential)]
       round: 0.00 [+- 0.00], round.block: 0.00 [+- 0.00], round.gc: 0.00 [+- 0.00], GC.calls: 6, GC.time: 0.01, time.total: 0.42, time.warmup: 0.00, time.bench: 0.42
      Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.816 sec - in org.roaringbitmap.TestBenchmarkIterator

lemire:
@bobymicroby  Please see my benchmarking code there: 

https://github.com/lemire/RoaringBitmap/blob/master/src/test/java/org/roaringbitmap/TestBenchmarkIterator.java

lemire:
@bobymicroby 

You should do fine with getCardinality(), but it should only be a few cycles faster getSizeInBytes(). The time you save, if any, should be dwarfed by the rest of the computation. A single cache fault can have a larger effect...

lemire:
@bobymicroby 

I have created an issue following this PR and my merge:

https://github.com/lemire/RoaringBitmap/issues/20

bobymicroby:
Most benchmarking tools beside JMH tend to give very faulty and misleading results.  I have observed better throughput both in the micro-benchmark and in real scenario. GC will always stab you in the back when you generate many Iterator objects ( I am not quite sure that you achieve this in your benchmark )  . 

I really would suggest that you try to do a transition to JMH 

Shorts slides about what pitfalls are JMH helping to avoid : http://www.oracle.com/technetwork/java/jvmls2013ship-2014087.pdf

This fine collection of resources about JMH http://psy-lob-saw.blogspot.com/p/jmh-related-posts.html  can help as well.

PR #22, opened 2015-03-28
Optimize AND

gssiyankai:
This PR optimizes AND (in-place and not-in-place) on roaring bitmaps.

So far, the AND operation between a bitmap A and a bitmap B used to be linear with n where:
  - n is the number of chunks with a key smaller than k
  - k is the min of kA and kB
  - kA is the max chunk key of A
  - kB is the max chunk key of B

Now, the AND operation is linear with N where:
  - N is the number of chunks with keys that are in both A and B

To have an idea of the improvements, I've set up some JMH benchmarks. Here are my results before and after my changes:
```
BEFORE
Benchmark                          Mode  Cnt   Score   Error  Units
RoaringBitmapBenchmark.and         avgt   10  58.256 ± 0.783  us/op
RoaringBitmapBenchmark.inplaceAnd  avgt   10   4.068 ± 0.411  ns/op

AFTER
Benchmark                          Mode  Cnt  Score   Error  Units
RoaringBitmapBenchmark.and         avgt   10  1.614 ± 0.026  us/op
RoaringBitmapBenchmark.inplaceAnd  avgt   10  3.869 ± 0.242  ns/op
```

To launch the JMH benchmarks locally, have a look at the README.

Let me know if this looks good to you and I will add the same optimization to the memory-mapped roaring bitmaps.


bobymicroby:
@gssiyankai Awesome results . Cheers :)

lemire:
@gssiyankai 

Thank you. I will have a look soon.

lemire:
@gssiyankai  

Thanks. This is neat code. It is well written.

You have established that, in some cases, your new code could be much faster. Congratulations!

I will merge this one way or another, but maybe we can take some time to assess it properly.

1. I would invite you to temporarily give distinct names to your function. Let us call them "newand" or something. We could then assess them in one run without the pain of having to recompile. Yes, this means that some clean-up is required prior to any new release... but I think it might be worth the effort.

2. You benchmark on a best-possible case for this PR. That is fine, but imagine that you have a more mundane set of bitmaps, where, say, you intersect a bitmap with itself. Or imagine a bitmap having containers at keys 1,3,5,... and another at keys 2,4,6,... In such cases, the binary search will prove counterproductive. Would the performance degrade? If so, by how much?

Another case of interest is when one bitmap is tiny (with one container) and the other one is large (containing many containers). This is clearly a case where your approach ought to help, and the current code has a TODO on this very issue: 

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L58


I think we need to know about the performance degradation. It is nice to be 30x faster in some fabricated cases, but less nice if you are 2x slower in other common cases. If there are performance degradations, we need to know about them so we can route around it...


To see what I have in mind... note that we already use intersections based on binary search...

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/Util.java#L309

But we use an approach based on binary search (called one-sided galloping) only when one of the two inputs is much smaller than the other one.

lemire:
I also do think we want a straight binary search that returns negative indexes. I think we want something closer to advanceUntil:

https://github.com/lemire/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/Util.java#L29

gssiyankai:
@lemire 

Thanks for the review. I submit some changes following your comments.

Below are the benchmarks results:
```
Benchmark                                                Mode  Cnt    Score   Error  Units
o.r.a.n.bestcase.RoaringBitmapBenchmark.and              avgt   10   57.771 ± 0.569  us/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_and      avgt   10    3.598 ± 0.097  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_newand   avgt   10    3.620 ± 0.072  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.newand           avgt   10    1.480 ± 0.025  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.and             avgt   10  319.256 ± 4.065  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_and     avgt   10    4.184 ± 0.232  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_newand  avgt   10    4.070 ± 0.162  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.newand          avgt   10  315.622 ± 6.262  us/op
```
As you can see, the new and is as performant as the old one even in the worst case scenario (I did not expect that to be honest)

gssiyankai:
@lemire

Thanks for the review. I submitted some changes following your comments.

Below are the benchmarks results with the latest Oracle JDKs:
```
$ java -version
java version "1.7.0_76"
Java(TM) SE Runtime Environment (build 1.7.0_76-b13)
Java HotSpot(TM) 64-Bit Server VM (build 24.76-b04, mixed mode)

Benchmark                                                Mode  Cnt    Score   Error  Units
o.r.a.n.bestcase.RoaringBitmapBenchmark.and              avgt   10   59.210 ± 1.371  us/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_and      avgt   10    3.864 ± 0.227  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_newand   avgt   10    3.786 ± 0.451  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.newand           avgt   10    1.600 ± 0.025  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.and             avgt   10  329.371 ± 4.950  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_and     avgt   10    4.098 ± 0.297  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_newand  avgt   10    4.077 ± 0.142  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.newand          avgt   10  322.263 ± 4.875  us/op
```
```
$java -version
java version "1.8.0_40"
Java(TM) SE Runtime Environment (build 1.8.0_40-b25)
Java HotSpot(TM) 64-Bit Server VM (build 25.40-b25, mixed mode)

Benchmark                                                Mode  Cnt    Score   Error  Units
o.r.a.n.bestcase.RoaringBitmapBenchmark.and              avgt   10   81.759 ± 1.174  us/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_and      avgt   10    1.423 ± 0.037  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.inplace_newand   avgt   10    1.430 ± 0.044  ns/op
o.r.a.n.bestcase.RoaringBitmapBenchmark.newand           avgt   10    1.414 ± 0.013  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.and             avgt   10  256.888 ± 8.964  us/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_and     avgt   10    1.436 ± 0.043  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.inplace_newand  avgt   10    1.437 ± 0.056  ns/op
o.r.a.n.worstcase.RoaringBitmapBenchmark.newand          avgt   10  226.606 ± 1.869  us/op
```

As you can see, the new and is as fast or faster than the old one, even in the worst case scenario (I did not expect that to be honest).

Let me know if I missed something.

lemire:
Thank you, I will review this shortly.

lemire:
Code looks good to me. Merging. 

I will review further and stress test.

lemire:
Please hold one while I review the code more carefully.

lemire:
Now that the PR has been merged, I have turned this into an issue: https://github.com/lemire/RoaringBitmap/issues/23

c.c. @gssiyankai  @bobymicroby

PR #25, opened 2015-04-02
Benchmarks

bobymicroby:
@lemire I have translated the benchmarks as-is . Please review them and if you merge, you can get rid of the old ones as well.


lemire:
Looks good. Merging.

PR #28, opened 2015-05-20
Implementing the checkAdd method

samytto:
This code intends to provide a new add method to RoaringBitmap, which will return a boolean after inserting an int, indicating if the inserted int was or wasn't already in the set before insertion. Tests and JMH benchmarks for this method were also implemented. 

Results show that the addCheck method was about 1 milliseconds slower than the precedent one.

Same thing for remove method will fellow.

Greetings!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,14 @@
 language: java
 
 jdk:
+  - oraclejdk8
   - oraclejdk7
   - openjdk7
-  - openjdk6
 
 install: true
+
+branches:
+  only:
+    - master
 
 script: mvn test
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 upcoming
  - faster intersections (and) in some cases thanks to @gssiyankai
+ - fast range add and remove
+ - improved select performance (Bob Potter)
+ - serialization and deserialization are faster (deserialization is about 30% less CPU intensive)
 
 version 0.4.9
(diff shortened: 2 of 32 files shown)


lemire:
@samytto 

Thanks for the PR.

Can you please make sure that you only add the lines that need to be added so I can review the proposed changes safely? For example, the current PR modifies unrelated files like ArrayContainer.java. You should only modify what needs to be modified. One needs to examine and understand all changes. Note that according to Travis, the PR breaks the unit tests: ideally, we would prefer not to have such problems.

If you prefer I can do a manual code integration, picking only the relevant functions.

lemire:
@samytto It might easier to withdraw this PR and re-issue a new one starting from a clean fork of the main code.

samytto:
Ok. It is mixed with the changes of an older pull request. I will clean that.

lemire:
@samytto  Excellent. Looking forward to it.

As I said, it might be easier to start anew with a fresh fork and a new PR.

lemire:
I have manually merged the PR : 

https://github.com/lemire/RoaringBitmap/commit/3d24f5785ca8ddd4bc856fa2ac9098b23e5fed4e

To avoid the need for manual mergers in the future, please sync your code with the master branch and only modify what needs to be modified.

samytto:
Ok. Thanks. I have checked, and the new parts have been well merged with the actual code.

Cheers!

PR #30, opened 2015-06-10
Resolving important bugs that could occur during logical operations

samytto:
Hi,

This pull request intends to resolve important bugs that could be encoutered when performing logical operations. In fact, we should compare using keys represented by 32-bit integers than shorts. With the latter, as Java doesn't deal with unsigned integers, values that use all represantative 16 bits, as 65530, will became negative values if represented by short keys, and Java will always consider such values smaller than all positive values which are lower than their unsigned values. 

Without this update, the actual code could return Roaring bitmaps with non sorted first level arrays after logical operations, making binary searches unuseful on the first level.

Cheers!


gssiyankai:
@samytto,

Thanks for spotting the issue. I've submitted some tests and fixed some missing operator methods @db81d76282e69c34febfb2feaa7f86e43242fc5b.
Can you merge it in your branch so that it will be included in the PR ?

I was also thinking about making *getKeyAtIndex* return an int. Calls to this method are always combined with *toIntUnsigned*. It would simplify the code and avoid falling into the problem again.
What do you think ? (cc @lemire )

lemire:
@gssiyankai 

Sounds reasonable.

lemire:
@gssiyankai @samytto 

I am adding tests of my own since Samy does not provide tests. It seems that we do have a problem, but I am not yet clear on whether Samy's approach is the right fix.

gssiyankai:
@lemire,

You can have a look at the tests I submitted to cover the problem reported by @samytto : db81d76282e69c34febfb2feaa7f86e43242fc5b

HTH

lemire:
Ok. I am investigating.

samytto:
@gssiyankai 

Thanks for having added the missing calls to shortTointUnsigned method. 

About making getKeyAtIndex return an int, I think that it shouldn't not be a well choice, because in many parts in the code, we need to retrieve a key in the short format (see RoaringBitmap.select(int)). Returning an int in such cases, will push us to do a suplementary cast form int to short; adding an unnecessary work that sould be avoided.

However, letting the getKeyAtIndex method as it is and converting from short to unsigned int when needed, will not add any suplementary work. 

Your tests seems ok for me. However, I think that it should be better to test using many more representative values, in lieu of only 1 and (1<<31).

Cheers!

lemire:
@samytto @gssiyankai @owenkaser 

Note that we have just defined the behaviour of Roaring for "large" (or signed) integers. See my remark : 

https://github.com/lemire/RoaringBitmap/commit/7eb340a751e2caaafffd41529d724022e840441e



@gssiyankai  Since you have worked on this, and I do not quite understand the interaction between your work and Samy's work, I have assigned the PR to you. Please merge it when you think it is ok.

samytto:
I have merged @gssiyankai code and added some tests for logical operations with big ints.

gssiyankai:
@samytto,
I pushed a new util method to compare unsigned shorts: 0414cb7 to keep working with shorts.
This approach fixes the problem you spotted, is not more costly than yours and impacts less code.
I am merging and closing the PR.
Revert if you disagree.

PR #33, opened 2015-07-07
Some (naive) SparseBitSet benchmarks for comparison

sabi0:
Upgraded to the latest JMH (1.10.1) and added a SparseBitSet construction benchmark


lemire:
Excellent.

PR #37, opened 2015-07-30
Fixing a real bug

samytto:


lemire:
Thanks. This function is indeed buggy and it will be removed in the next revision, please see issue 32 

https://github.com/lemire/RoaringBitmap/issues/32

I will merge your PR for now, but do not rely on this function being present in future versions.

samytto:
Awesome!

Cheers!

PR #43, opened 2015-09-20
Make ImmutableRoaringArray thread-safe for read access

gpunti:
When calling contains() method from different threads an IllegalStateException was raised. This can be checked by running the provided unit test with the previous version of the code.

By using a duplicate buffer instead of the global one, the problem is fixed ( this is the same that the serialize method does )


lemire:
Good catch.

Issue #36, opened 2015-07-30
Feature Request: Set operations (does_it_intersects, or_cardinality, and_cardinality...)


llogiq:
It should be possible to implement fairly fast set operations like `intersects(RoaringBitmap)` (a method that even `BitSet` has) and `containsAll(RoaringBitmap)` (a method that `BitSet` lacks, but could be quite handy in some cases).

Also a fast method to *count* the bits of intersections/unions might be useful (I'm using RoaringBitmaps for a Jaccard Index, where both is needed – so a method that does both fast would be even more awesome). Granted, the latter is really specific, so I have no hard feelings if it doesn't get implemented.

Perhaps providing a method `Stream<Pair> pairStream(RoaringBitmap, RoaringBitmap)` where each `Pair` has an index + two longs with the bits wherever at least one of the bitmaps has bits could be used to solve all of the above issues in a generic way?

lemire:
> It should be possible to implement fairly fast set operations like intersects(RoaringBitmap) (a method that even BitSet has) 

You are correct. This should be added.

> and containsAll(RoaringBitmap) (a method that BitSet lacks, but could be quite handy in some cases)

This could be added for convenience, but it can be implemented as ```(RoaringBitmap.xor(first,second).getCardinality() == 0)```.

> Also a fast method to count the bits of intersections/unions might be useful

Agreed. This should be added as well.

> Perhaps providing a method Stream<Pair> pairStream(RoaringBitmap, RoaringBitmap) where each Pair has an index + two longs with the bits wherever at least one of the bitmaps has bits could be used to solve all of the above issues in a generic way?

I do not understand this proposal. Can you give an example?

llogiq:
> `(RoaringBitmap.xor(first,second).getCardinality() == 0)`

This requires potentially a good amount of memory, whereas a direct method should be fairly frugal in that regard. I'm unsure how this would affect performance.

> I do not understand this proposal. Can you give an example?

Yea, I was thinking in terms of BitSets. It doesn't make sense anywhere else. Disregard this one.

lemire:
By the way... forget my xor thing, it won't do the job.

You need andNot...

lemire:
This should probably be a priority for future work.

c.c. @llogiq  @gpunti @gssiyankai @rorygraves

lemire:
@llogiq 

Version 0.5.6 implements "intersects" which is the "does_it_intersects" part of this issue.

[...]

PR #52, opened 2015-11-25
#36 add andCardinality method that does not modify input RoaringBitSet

JervenBolleman:
As I have a similar issue as #36 with requiring a lot of intersect counts and wishing to avoid the cost of allocation I implemented the basic andCardinality method.

The logic is shamelessly stolen from the and method. The only difference is that only an int is incremented instead of a new RoaringBitSet build as are the two test cases.

### DIFF
--- a/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -87,15 +87,12 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
      * modified. This operation is thread-safe as long as the provided
      * bitmaps remain unchanged.
      *
-     * If you have more than 2 bitmaps, consider using the
-     * FastAggregation class.
-     *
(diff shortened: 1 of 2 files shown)


lemire:
Good first start.

lemire:
This will be part of release 0.5.12 (coming out today).

lemire:
Note that we still materialize the containers... but it should still be faster. So good work.

JervenBolleman:
Thank you. I will have a look at adding the orCardinality as well. 
And then have a good look at the underlying data structure to see if there is an even faster possible solution.

lemire:
orCardinality has been added.

JervenBolleman:
@lemire thank you. I have code on my local branch that does the andCardinality without materializing the Containers. I just want to improve the tests cases to make sure I touch all options before asking for a merge. This is quite a bit more code again still based on your existing code, but more spaces for a mistake to break the code subtly.

lemire:
Excellent.

Issue #53, opened 2015-11-26
ClassCastException on cloning ImmutableRoaringReverseIntIterator


navis:
in `ImmutableRoaringReverseIntIterator.clone()`,
```
public IntIterator clone() {
    try {
        ImmutableRoaringIntIterator x = (ImmutableRoaringIntIterator) super.clone();
        x.iter = this.iter.clone();
        x.cp = this.cp.clone();
        return x;
    } catch (CloneNotSupportedException e) {
        return null;// will not happen
    }
}
```
super.clone() is not ImmutableRoaringIntIterator, making CCE.

lemire:
Will issue a release today.

Closing for now.

PR #54, opened 2015-11-26
fix for #53 ClassCastException on cloning ImmutableRoaringReverseIntIterator

navis:
trivial fix


lemire:
Good catch.

lemire:
I will issue a release right now.

PR #57, opened 2015-12-04
#36 andCardinality methods added to the Containers avoiding materialization

JervenBolleman:
Hi this is the code to do the andCardinality without materializing the Containers. Which should be even faster. Again my apologies for messing up the code formatting. Hope this is useful.Before pulling could you have a look at the test cases to see if they make sense. I pulled them out of the or tests and added all against all container tests. Reading from the comments I think it does  a decent job of covering most possibilities.

### DIFF
--- a/src/main/java/org/roaringbitmap/BitmapContainer.java
+++ b/src/main/java/org/roaringbitmap/BitmapContainer.java
@@ -232,16 +232,7 @@ public final class BitmapContainer extends Container implements Cloneable {
             newCardinality += Long.bitCount(this.bitmap[k]
                     & value2.bitmap[k]);
         }
-        if (newCardinality > ArrayContainer.DEFAULT_MAX_SIZE) {
-            final BitmapContainer answer = new BitmapContainer();
-            for (int k = 0; k < answer.bitmap.length; ++k) {
-                answer.bitmap[k] = this.bitmap[k]
-                        & value2.bitmap[k];
-            }
-            return newCardinality;
-        } else {
-        	return newCardinality;
-        }
+        return newCardinality;
     }
     
     @Override
--- a/src/main/java/org/roaringbitmap/BitmapContainer.java
+++ b/src/main/java/org/roaringbitmap/BitmapContainer.java
@@ -191,6 +191,18 @@ public final class BitmapContainer extends Container implements Cloneable {
         }
         return answer;
     }
+    
+    @Override
+    public int andCardinality(final ArrayContainer value2) {
+        int answer=0;
+        int c = value2.cardinality;
+        for (int k = 0; k < c; ++k) {
+            short v = value2.content[k];
+            if (this.contains(v))
+                answer++;
+        }
+        return answer;
+    }
 
     @Override
     public Container and(final BitmapContainer value2) {
@@ -213,7 +225,25 @@ public final class BitmapContainer extends Container implements Cloneable {
         ac.cardinality = newCardinality;
         return ac;
     }
-
+    @Override
+    public int andCardinality(final BitmapContainer value2) {
+        int newCardinality = 0;
+        for (int k = 0; k < this.bitmap.length; ++k) {
+            newCardinality += Long.bitCount(this.bitmap[k]
+                    & value2.bitmap[k]);
+        }
+        if (newCardinality > ArrayContainer.DEFAULT_MAX_SIZE) {
+            final BitmapContainer answer = new BitmapContainer();
+            for (int k = 0; k < answer.bitmap.length; ++k) {
+                answer.bitmap[k] = this.bitmap[k]
+                        & value2.bitmap[k];
+            }
+            return newCardinality;
+        } else {
+        	return newCardinality;
+        }
+    }
+    
     @Override
     public Container andNot(final ArrayContainer value2) {
         final BitmapContainer answer = clone();
@@ -1054,6 +1084,11 @@ public final class BitmapContainer extends Container implements Cloneable {
     @Override
     public Container and(RunContainer x) {
         return x.and(this);
+    }
+    
+    @Override
+    public int andCardinality(RunContainer x) {
+        return x.andCardinality(this);
     }
 
     @Override
--- a/src/main/java/org/roaringbitmap/Container.java
+++ b/src/main/java/org/roaringbitmap/Container.java
@@ -99,6 +99,32 @@ public abstract class Container implements Iterable<Short>, Cloneable, Externali
     }
 
     /**
+     * Computes the bitwise AND of this container with another
+     * (intersection). This container as well as the provided container are
+     * left unaffected.
+     *
+     * @param x other container
+     * @return aggregated container
+     */
+    public int andCardinality(Container x) {
+    	if (this.getCardinality() == 0)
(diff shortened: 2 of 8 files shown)


lemire:
Excellent. I will review shortly.

src/main/java/org/roaringbitmap/BitmapContainer.java
```diff
@@ -213,7 +225,25 @@ public Container and(final BitmapContainer value2) {
...
     }
-
+    @Override
+    public int andCardinality(final BitmapContainer value2) {
+        int newCardinality = 0;
+        for (int k = 0; k < this.bitmap.length; ++k) {
+            newCardinality += Long.bitCount(this.bitmap[k]
+                    & value2.bitmap[k]);
+        }
+        if (newCardinality > ArrayContainer.DEFAULT_MAX_SIZE) {
```
gssiyankai: I think you can just return `newCardinality` here.

src/main/java/org/roaringbitmap/Container.java
```diff
@@ -99,6 +99,32 @@ else if (x instanceof BitmapContainer)
...
+     */
+    public int andCardinality(Container x) {
+    	if (this.getCardinality() == 0)
+    		return 0;
+    	else if (x.getCardinality() ==0)
+    		return 0;
+		else
+		{
+    		if (x instanceof ArrayContainer)
+            	return and((ArrayContainer) x).getCardinality();
```
gssiyankai: `andCardinality(Container)` should be called here instead of `and(Container).getCardinality()`

src/main/java/org/roaringbitmap/RoaringBitmap.java
```diff
@@ -104,9 +104,8 @@ public static int andCardinality(final RoaringBitmap x1,
             if (s1 == s2) {
                 final Container c1 = x1.highLowContainer.getContainerAtIndex(pos1);
                 final Container c2 = x2.highLowContainer.getContainerAtIndex(pos2);
-                final Container c = c1.and(c2);
                 // TODO: could be made faster if we did not have to materialize container
-                answer += c.getCardinality();
+                answer += c1.andCardinality(c2);
```
gssiyankai: :+1:

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -459,7 +459,33 @@ public Container and(ArrayContainer x) {
         return ac;
     }
 
-
+    @Override
+    public int andCardinality(ArrayContainer x) {
```
gssiyankai: LGTM

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -496,6 +522,38 @@ public Container and(BitmapContainer x) {
...
     }
 
+    @Override
+    public int andCardinality(BitmapContainer x) {
+        // could be implemented as return toBitmapOrArrayContainer().iand(x);
+        int card = this.getCardinality();
+        int cardinality=0;
+        if (card <=  ArrayContainer.DEFAULT_MAX_SIZE) {
+            // result can only be an array (assuming that we never make a RunContainer)
+            if(card > x.cardinality) card = x.cardinality;
```
gssiyankai: This line can be removed as `card` is not used afterwards.

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -496,6 +522,38 @@ public Container and(BitmapContainer x) {
...
+            cardinality=0;
```
gssiyankai: This line is not necessary as `cardinality` has already been initialized before and not modified since.

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -496,6 +522,38 @@ public Container and(BitmapContainer x) {
...
+                for (int runValue = runStart; runValue <= runEnd; ++runValue) {
+                    if ( x.contains((short) runValue)) {// it looks like contains() should be cheap enough if accessed sequentially
+                        cardinality++;
+                    }
+                }
+            }
+            return cardinality;
+        }
+        // we expect the answer to be a bitmap (if we are lucky)
+        BitmapContainer answer = x.clone();
```
gssiyankai: I believe it will be simpler and more efficient (because the bitmap container `x` is cloned here) to always use the algorithm in the if block above.

gssiyankai:
Good job @JervenBolleman.
I added some notes on your changes.
HTH

src/main/java/org/roaringbitmap/RunContainer.java
JervenBolleman: @gssiyankai what does LGTM mean?

JervenBolleman:
Ok I will @gssiyankai & @lemire  I will adapt the code to take into account the comments.

src/main/java/org/roaringbitmap/RunContainer.java
gssiyankai: Looks Good To Me

src/main/java/org/roaringbitmap/RunContainer.java
JervenBolleman: That makes sense and doings so leads to no test regressions.

JervenBolleman:
I am regressing code wise on two methods the RunContainer.andCardinality(RunContainer x) as my code was broken and the ArrayContainer.andCardinality(ArrayContainer x).

JervenBolleman:
Ok, I managed to make it work for ArrayContainer.andCardinality(ArrayContainer x). Duplicating a lot of code in the Util class.

The RunContainer one is not yet implemented.

JervenBolleman:
Having finally understood the RunContainer datastructure the andContainer method became obvious. Still a lot of code duplication but until java8 becomes an option I don't see a way to reduce the duplication.

lemire:
Thanks.

It is probably best to let  @gssiyankai review the changes and green light them before considering a merge.

gssiyankai:
@lemire,
LGTM. Let's merge.

lemire:
Ok. This is now merged. Thanks for the excellent work. I'll be adding @JervenBolleman's name to the list of contributors later today.

We still have work ahead however...

This PR does not cover the buffer package. So a next step would be to port it over so that users of the buffer package can also benefit from it. Presumably, that is not too difficult.

And, of course, we also need to complete the orCardinality optimization.

lemire:
Note to @JervenBolleman : for extra points, you could produce a PR which would extend your changes to the buffer package. It would mostly just be copy-and-paste.

(Yes, we have to duplicate the logic throughout which sounds very bad, but that's necessary for performance reasons.)

JervenBolleman:
@lemire, @gssiyankai, @samytto  thanks for accepting. I suspect that in java 8 we could get the same code doing more by passing functions into shared loop constructs. I suspect that will JIT to the same type of assembly on a recent 8 jvm or at least the Graal stuff is quite advanced in inlining functions wholesale. 

But for now I need to do the same for andNot and or and move it into the buffer package as well. This was the code that is on the hotpath at work I won't have time for that until the x-mas holiday.

PR #70, opened 2016-01-27
Add functionality to transform BitSet to RoaringBitmap efficiently

okrische:
This tries to bypass any add() function, where one has to decompress or
re-size arrays, etc..

Could still be improved in the loops. Also the decision between RunContainer/ArrayContainer/BitmapContainer needs some work.


lemire:
Looks like a good idea. Will review.

c.c. @gssiyankai

gssiyankai:
@okrische 
Good job.
I just added some comments on the changes.
As you mentionned, BitmapContainer are not used at the moment. I would recommand to support this transformation before merging.

lemire:
One thing that would be very useful, if you have time, is to do a benchmark, if only an informal one, to check how fast things run compared to simply iterating over the BitSet and calling "add" to a RoaringBitmap. It would be best if such a benchmark included varied BitSet objects (small, large, dense, sparse). We would like to make sure that your function is always faster than the naive way.

Optional but nice: If you really want to be helpful, you could add a JMH benchmark to our set. I think that @gssiyankai could give pointers as he is very good at that.


Regarding the code, this is a good starting point. I would suggest simplifying/optimizing the code :

Assuming that the BitSet is long enough (has >= 1024 64-bit words):
* Start by calling toLongArray on the BitSet .
* Iterate over chunks as is done now.
* Create a BitmapContainer for each chunk. Call computeCardinality. Next check cardinality. If too low, call toArrayContainer.
* Don't worry about run containers. Just have the user call runOptimize on the final result if desired. This will simplify your logic as you won't need to worry about runs.

You need a special code path for the "last chunk" (a chunk that does not span 1024 words in the original BitSet) and for BitSet that span fewer than 1024 words. I think you want to first determine the cardinality. To do so you may need a special function, but please don't iterate over the set bits to computer the cardinality. Instead use ``Long.bitCount``.

okrische:
I like the idea of skipping RunContainer at all. And sure, i can use Bitset#toLongArray and use it directly in the BitmapContainer. 

Thanks to both of you for your comments.

gssiyankai:
@okrische 

To add a new JMH benchmark, you would need to : 
- Add a new class under the jmh module
- Annotate the class with 
```java
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
```
- Add a setup method where you construct the bitset to convert. Annotate it with:
```java
@Setup
```
- Add a method where you convert the bitset to a roaring bitmap. Annotate it with:
```java
@Benchmark
```
You can have a look at the existing benchmarks (this one for example can be a good starting point: https://github.com/RoaringBitmap/RoaringBitmap/blob/master/jmh/src/main/java/org/roaringbitmap/BasicBenchmark.java)

To run the benchmark, you can either add a main method in your class and launch it from your IDE, or use the script run.sh. Make sure to give the name of the class you added (otherwise, all of the benchmarks will be launched)

okrische:
Using long[] simplified a lot. 

Also added BitSetUtilBenchmark in jmh.

I still have some doubts about efficiently iterating over one 64bit word to extract the bits for the ArrayContainer, though. But it was fun to think about it, soooo long ago to think about such low level things.

I kept the BitSet transformation, as i can use a shortcut for cardinality < 4096, as i do not need to do this extra toLongArray() copy.

Since its still java 1.6, i also added a toLongArray() implementation for BitSet, well, using the javadoc from BitSet in 1.8, in the way they suggested it. When you guys change to java 1.7, this can be removed and speeds up the whole transformation again.

lemire:
@gssiyankai  

I checked and RoaringBitmap *is* compatible still with Java 6. Note however that we have not systematic testing (travis) in place.

lemire:
@okrische

I'll review the code soon. Thanks.

gssiyankai:
@lemire,

Great. If so, I suggest to add it back in travis CI (we used to have it
before if I'm correct).
I'll submit the change if you agree.

lemire:
@gssiyankai  

I don't object to you adding it, but there is an opportunity here for upgrading to Java 7.

Thoughts?

gssiyankai:
Ah, ok.

IMHO, it's reasonable to upgrade. Java 7 offers a richer concurrency api
that might be helpful to parallelize executions (and we know that Roaring
internal structure is well suited for such things).

Plus I'm not sure we have a lot of Java 6 users. The end of public updates
for Oracle Java 6 was on Feb 2013 (the end of extended support is scheduled
on Dec 2018, but this case should be very rare). Besides, as you mentionned
before, future versions of Android will be based on OpenJdk and will be
Java 8 compatible.

I would just point out that the upgrade to java 7 should be made available
in a major release of Roaring(version 0.6.xxx) so that we can still provide
bug fixes for java 6.

How about upgrading to java 8 directly ?

okrische:
I am of course in favour of java 8. But not because of the concurrency stuff (i personally do not like any executor/runnable things in libraries), but rather because of lambdas. For example, one could stream efficiently indexes into a bitmap via an IntStream. Or like in other threads, offering nice collectors for OR/AND operations between bitmaps (and then, by choice, allowing parallel streaming, but externally, not from within the library)

lemire:
I propose moving the java6/java8 discussion to an actual issue I just created. 

https://github.com/RoaringBitmap/RoaringBitmap/issues/74

lemire:
@okrische 

>  When you guys change to java 1.7, this can be removed and speeds up the whole transformation again.

I think there there is agreement to move to 1.7 as major vendors have upgraded to 1.7. So starting with version 0.6, RoaringBitmap will no longer be compelled to support Java 1.6. This means that we can merge your PR without sacrificing performance.

However, this is currently stalled because travis is failing for us... I have opened an issue...

https://github.com/RoaringBitmap/RoaringBitmap/issues/75

I think that @gssiyankai  has agreed to look at it. 

Once the build is officially passing the tests, we shall be able to merge in your PR.

Thanks for your patience.

lemire:
Thanks to @gssiyankai , we are back in business.

lemire:
@okrische  

I am working on a PR that I will submit to your fork soon.

lemire:
@okrische  Please see...

https://github.com/okrische/RoaringBitmap/pull/1

Review and possibly merge  my PR. Then, possibly, we can merge back in the main RoaringBitmap.

okrische:
Yay! With your changes, the conversion is now 4 times faster according to jmh with my dataset, where i have only a max of 96 bits per BitSet ( so it uses only ArrayContainers)

```bash
$:~/git/RoaringBitmap/jmh$ ./run.sh org.roaringbitmap.BitSetUtilBenchmark
...
Benchmark                                            Mode  Cnt       Score       Error  Units
BitSetUtilBenchmark.BitSetToRoaringByAddingBitByBit  avgt    5  321265.672 ± 16719.292  us/op
BitSetUtilBenchmark.BitSetToRoaringUsingBitSetUtil   avgt    5   80315.676 ±   711.213  us/op
```

If i duplicate each bitset a hundred times to the right (96000 bits max), it now even gets 10 times faster.

```bash
...
Benchmark                                            Mode  Cnt        Score        Error  Units
BitSetUtilBenchmark.BitSetToRoaringByAddingBitByBit  avgt    5  9869858.180 ± 129760.451  us/op
BitSetUtilBenchmark.BitSetToRoaringUsingBitSetUtil   avgt    5  1060313.473 ±  10476.022  us/op
```

Not to mention memory usage...

Thanks!

lemire:
@okrische 

Thanks for your hard work.

lemire:
@okrische 

Your name was added...

https://github.com/RoaringBitmap/RoaringBitmap/commit/4306038ea4487d28410bd9a814ae1c1a7a23b7c5

PR #71, opened 2016-01-28
fixed bug on hashCode() method. It entered an infinite loop.

gpunti:
fixed: The container pointer wasn't advanced, and when adding an instance to a Map the application hanged.

The equals() method should also be implemented to be consistent with the hashCode(). I don't know exactly when two arrays should be considered equal ( checking every single byte might be too expensive, depending on how the arrays are used )


lemire:
Merging.

If you have concerns regarding equals, please issue another issue.

lemire:
See release 0.5.19.

lemire:
Much appreciated

lemire:
Note that hashCode takes linear time, which means that the whole bitmap must be scanned.

PR #73, opened 2016-01-29
fix: java.lang.ClassCastException

gpunti:
org.roaringbitmap.buffer.ImmutableRoaringBitmap cannot be cast to org.roaringbitmap.buffer.MutableRoaringBitmap


lemire:
Thanks.

lemire:
Fixed in release 0.5.20.

Thanks.

lemire:
@gpunti 

Note that your membership to the RoaringBitmap organization is currently private: https://github.com/orgs/RoaringBitmap/people/gpunti

Given your valued contribution, you might want to make it public.

gpunti:
Made public. I hadn't noticed it was private. Thanks.

PR #96, opened 2016-03-25
Adding ranged or/and/xor/andnot operations

kishoreg:


lemire:
Looks promising to me...

* There seems to be a minor style infraction...
```
/home/travis/build/RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java:454: 'if' construct must use '{}'s. [NeedBraces]
```
I could fix it easily enough, but I don't want to merge code the fails our style guideline.

(What we would really need is a way for maven to reformat the code. Sadly, all I could find is a tool to check the style, so the reformatting must still be done manually by the programmer.)

* Can you port your API extension to ``RoaringBitmap``? We try to keep the buffer and standard Roaring instances in sync as far as API is concerned. Mostly it involves copying and pasting with some minor API tweaks. If you are stuck I could do it too. But I expect you will find it easy. Make sure to carry over the testing as well.

kishoreg:
The api's should work for RoaringBitmap as well. I could not see a place to
put the api's that worked for RoaringBitmap and ImmutableRoaringBitmap

lemire:
> The api's should work for RoaringBitmap as well. I could not see a place to
put the api's that worked for RoaringBitmap and ImmutableRoaringBitmap 

Maybe I misunderstand what you are saying. ``ImmutableRoaringBitmap.java`` and ``RoaringBitmap.java`` are very similar classes, but code changes made to ``ImmutableRoaringBitmap.java`` need to be copied to ``RoaringBitmap.java`` (with appropriate fixes). 

Yes, it is annoying, but it is only way we have to do it to support both memory-file mapping and regular use, without sacrificing performance. We need to have duplicated code.

If you don't want to do it, I could do it for you later. 

However, I'd rather you fix the style issue on your end, as it is only the addition of the curly brackets required by the Google style. It is only one line you need to fix.

kishoreg:
Added the missing remove method in iterator to make it 1.7 compatible.

lemire:
Looks good. I should merge this soon.

PR #98, opened 2016-03-26
Fixing selectRangeWithoutCopy

kishoreg:
The bug was in selectRangeWithoutCopy. The real bug is in  MappeableArrayContainer.remove(begin.end) method when begin=end=0. It removed everything in the container. It worked for cases where begin != end.

I worked around this by not invoking remove if lbStart == 0. I added the same check for lbLast as well though I dint check the behavior of MappeableArrayContainer.remove when lbLast == BufferUtil.maxLowBitAsInteger()

Test cases pass now. I also added additional test for selectRangeWithoutCopy.

### DIFF
--- a/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
@@ -443,7 +443,7 @@ public class ImmutableRoaringBitmap
    * @return new bitmap
    */
 
-  private static MutableRoaringBitmap selectRangeWithoutCopy(ImmutableRoaringBitmap rb,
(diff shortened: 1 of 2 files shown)


kishoreg:
Its still failing. Might need to log more info when it fails.

kishoreg:
It does not fail on my box. I increased the iterations etc but it still passed. Not sure why its failing on travis. Can you give it a try, also can you log more info when the test fails. 

if (!expectedResultSet.equals(actualResultSet)) {
          System.out.println("ERROR " + "rangeStart:" + rangeStart + "  rangeEnd:" + rangeEnd);
          System.out.println("rb1:" + rb1);
          System.out.println("rb2:" + rb2);
}

lemire:
I'm on it. 

I will actually revert your fix as you are correct, there is a bug higher up that needs fixing.

PR #100, opened 2016-04-02
Add test case to test unsignedIntersects() method

JunyuanZheng:
Hi, this is my first pull request, I will try to keep adding test cases to improve the coverage. If you have some suggestion please tell me.

Sincerely,

Junyuan Zheng


lemire:
Excellent!

lemire:
I'll merge this right away, as it can do no harm, but I would suggest you also cover BufferUtil/unsignedIntersects for symmetry.

lemire:
Added your name to the list of contributors.

JunyuanZheng:
Thanks! I will try to cover that!

Issue #101, opened 2016-04-02
BufferUtil/unsignedLocalIntersect2by2Cardinality() method always return 0


JunyuanZheng:
unsignedLocalIntersect2by2Cardinality() method internal **pos** variable never chenge, might be a bug?

lemire:
The code is a result of some copy and paste. The variable should be removed as it is leftover from other code where it serves to indicate where we write the intersection. Note that we should check whether the same thing happens elsewhere including in BufferUtil.

Want to issue a PR to clean this up?

lemire:
+1 for spotting this mistake

JunyuanZheng:
I think I need to remove the whole method right?

lemire:
Let me add a unit test first.

lemire:
@JunyuanZheng 

What you have uncovered is a genuine bug, but it escaped our unit tests up to now. I have now added tests that expose the bug.

I think we need to increment pos near this line...

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/buffer/BufferUtil.java#L718

Want to submit a PR?

lemire:
Note that if you delete the method, the code won't build anymore.

JunyuanZheng:
I will try to improve the coverage first. Need to study the algorithm first before changing the code.

JunyuanZheng:
I will fix the issue on next Pull Request.

lemire:
Excellent.

PR #102, opened 2016-04-03
Fix issue #101, coverage improvement

JunyuanZheng:
Fix issue #101 , improve coverage to 80%

### DIFF
--- a/src/main/java/org/roaringbitmap/Util.java
+++ b/src/main/java/org/roaringbitmap/Util.java
@@ -20,7 +20,7 @@ public final class Util {
    * return length. Based on code by O. Kaser.
    *
    * @param array array to search within
-   * @param pos starting position of the search
+   * @param pos starting position of th search
    * @param length length of the array to search
    * @param min minimum value
    * @return x greater than pos such that array[pos] is at least as large as min, pos is is equal to
@@ -200,7 +200,7 @@ public final class Util {
     }
     int firstword = start / 64;
     int endword = (end - 1) / 64;
-    bitmap[firstword] ^= ~(~0L << start);;
+    bitmap[firstword] ^= ~(~0L << start);
     for (int i = firstword; i < endword; i++) {
       bitmap[i] = ~bitmap[i];
     }
--- a/src/main/java/org/roaringbitmap/Util.java
+++ b/src/main/java/org/roaringbitmap/Util.java
@@ -20,7 +20,7 @@ public final class Util {
    * return length. Based on code by O. Kaser.
    *
    * @param array array to search within
-   * @param pos starting position of th search
+   * @param pos starting position of the search
    * @param length length of the array to search
    * @param min minimum value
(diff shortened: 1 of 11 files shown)


JunyuanZheng:
Right now the 80% coverage is the best that I can do. I am a beginner of software testing. I will spend some time in learning the code and then write new test cases.

src/main/java/org/roaringbitmap/Util.java
```diff
@@ -20,7 +20,7 @@
    * return length. Based on code by O. Kaser.
    *
    * @param array array to search within
-   * @param pos starting position of the search
+   * @param pos starting position of th search
```
gssiyankai: unintentional change I guess ...

src/test/java/org/roaringbitmap/buffer/TestUtil.java
```diff
@@ -0,0 +1,65 @@
...
+    public void testFillArrayANDNOT() {
+        LongBuffer data1 = LongBuffer.wrap(new long[]{1, 2, 4, 8, 16});
+        LongBuffer data2 = LongBuffer.wrap(new long[]{2, 1, 3, 7, 15});
+        LongBuffer data3 = LongBuffer.wrap(new long[]{2, 1, 3, 7});
+        short[] content = new short[5];
+        short[] content2 = new short[5];
+        short[] result = {0, 65, 130, 195, 260};
+        BufferUtil.fillArrayANDNOT(content, data1, data2);
+        Assert.assertTrue(Arrays.equals(content, result));
+        BufferUtil.fillArrayANDNOT(content2, data1, data3);
```
gssiyankai: I would move the last `BufferUtil.fillArrayANDNOT` in another test that would have the expected exception annotation, just to make the test more robust.
Indeed, if the first `BufferUtil.fillArrayANDNOT`is throwing the `IllegalArgumentException` because of a new regression, then the test will pass and the regression will go unnoticed.

gssiyankai:
@JunyuanZheng 
:+1:  Nice work!
I added some notes on your changes. Let me know if you need help.

JunyuanZheng:
I will fix the first one. What should I do for the second one, can you explain this to me, so I can fix that and submit a new PR. Thanks!

gssiyankai:
`testFillArrayANDNOT` is testing 2 things:
1. normal execution of `fillArrayANDNOT`
2. exception thrown by `fillArrayANDNOT` because of unsupported inputs
What I suggesting is to split these in 2 different tests. Is it more clear ?

JunyuanZheng:
Oh, yeah, that's right! I never think about that. Thanks! I will fix that later.

JunyuanZheng:
I made the change... If everything is good, please merge this pull request. Thanks!

gssiyankai:
Great!

lemire:
+1

lemire:
Released as 0.6.12.

Issue #105, opened 2016-04-11
Performance: RoaringIntIterator#advanceIfNeeded


okrische:
Right now the code looks like this:

```java
    @Override
    public void advanceIfNeeded(int minval) {
      while ((0xFFFF & (hs >>> 16)) < (0xFFFF & (minval >>> 16))) {
        ++pos;
        nextContainer();
      }
      if (hasNext()) {
        iter.advanceIfNeeded(Util.lowbits(minval));
        if (!iter.hasNext()) {
          ++pos;
          nextContainer();
        }
      }
    }
```

If i test it by this test:

```java
  @Test
  public void testFull() {
	  RoaringBitmap bitmap = new RoaringBitmap();
	  bitmap.add(0, Integer.MAX_VALUE);
	  PeekableIntIterator it = bitmap.getIntIterator();
	  it.advanceIfNeeded(Integer.MAX_VALUE-1);
  }
```

then i can see, that the above code has to iterate over 32766 containers to get the matching one.  RoaringBitmap already can do it much faster by using the same approach as in RoaringBitmap#contains

I originally did this in: * https://github.com/okrische/RoaringBitmap/commit/a67ae0f3a880435bfc67a00c971134fc52b004db

I try to adapt it to the current code.

lemire:
Indeed. The code could be faster in such cases.

PR #106, opened 2016-04-11
implement search for container, instead of iterating to it

okrische:
See https://github.com/RoaringBitmap/RoaringBitmap/issues/105

I added a private method "advanceIfNeededProposal" to RoaringBitmapIntIterator. So one just needs to switch.

Have no idea to have test both methods and compare, without making the iterator visible. 

I wrote a test, where one can see the difference, if switching from current to proposed method:

```java
  @Test
  public void testWithFullBitmap() {
	  bitmap.add(0, Integer.MAX_VALUE);
	  Instant start = Instant.now();
	  for(int i = 0; i < 10000; i++) {
		  PeekableIntIterator it = bitmap.getIntIterator();
		  it.advanceIfNeeded(Integer.MAX_VALUE-1);		  
	  }
	  System.err.println(Duration.between(start, Instant.now()));
  }
```

TestIterators ran through with the proposed methods.

### DIFF
--- a/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -4,7 +4,13 @@
 
 package org.roaringbitmap;
 
-import java.io.*;
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.Externalizable;
+import java.io.IOException;
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.io.Serializable;
 import java.util.Iterator;
 
 import org.roaringbitmap.buffer.ImmutableRoaringBitmap;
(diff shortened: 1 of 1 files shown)


okrische:
(ah, the test is java8, well, i did not add it to the code anyways, it was just for me to check locally)

lemire:
Thanks. Hopefully, I will have a look soon.

lemire:
Thank you.

Some thoughts:

1. Obviously, the current PR cannot be merged as it stands as it would just be "dead code". But you know this.

2. I don't understand why ``advanceIfNeededProposal`` has a separate code path for ``minval=0``. I also do not understand what it does. It seems to rewind the container which is not what the API would suggest (it says it advances if needed, not "rewinds or advance as needed"). In fact, ``advanceIfNeeded(0)`` should never do anything at all, I think.

3. Same goes for the rest of the function. It seems that sometimes you would rewind. For example, you do not check that ``highLowContainer.getIndex`` brings you were you need to be, that is, always beyond the point where you were, never before.

4. Are you sure that the method you want is ``advanceIfNeeded``. It almost feels like what you want to implement is something that is not currently supported by the API?

5. If you do want to accelerate ``advanceIfNeeded``,  I'd rather imagine that you would start from the existing ``advanceIfNeeded`` code, and make as small a change as possible. When you detect that you are in the wrong container, and that you need to go forward to an upcoming container (and only in that instance) you call ``highLowContainer.getIndex`` to quickly move to the right container, never going back. Since, as I understand it, this is a performance issue, I think we do need some kind of benchmark to see the effects of the optimizations, good or bad. We'd like to see the effect of the change on bitmaps made of just a couple or a handful of containers (to check that we are not regressing there), and then see the effect when there are many containers (to quantify the benefits). Note that since your PR would invariably make the code more complex (more lines of code), we do need some evidence that it improves performance, otherwise we are better off leaving the code as is.

c.c. @gssiyankai

okrische:
Hello,

thank you for your time.

2) When someone gives minval = 0, i wanted to reset it to the state as if the constructor was called. Maybe its not necessary.

3+4) You are right. When it says "advance", the iterator should always "advance" in one direction. I will fix that. Maybe a specific RandomAccessIntIterator makes the API more clear.

5) For my use case i needed an impl that avoids the iteration to find the container for minval, as i have bitmaps with hundreds of containers. So already the jump to the matching container relieved me from my urgent lookup problem.

But i agree, more testing is useful. For example, if minval is potentially in the current container, then one might not need to do another binarySearch for the matching container on another advanceIfNeeded call.

Yay, more work :>

okrische:
If you want, we can close this request. As there is more work to be done.

lemire:
<em>2) When someone gives minval = 0, i wanted to reset it to the state as if the constructor was called. Maybe its not necessary.</em>

This could be a good feature, but it is probably better to implement it through an intuitive API call.

<em>Maybe a specific RandomAccessIntIterator makes the API more clear.</em>

We may not need new classes per se, but we may need new functions.

<em>For my use case i needed an impl that avoids the iteration to find the container for minval, as i have bitmaps with hundreds of containers. So already the jump to the matching container relieved me from my urgent lookup problem.</em>

Yes, but performance issues can be counterintuitive... so while we may think that new code improves the performance, checking is still important.

lemire:
Ok. Let us close this temporarily.

PR #111, opened 2016-07-09
Checkstyle fixes

okrische:
Hey,

can someone please verify, if the current master branch can not be built because of checkstyle violations? Because here it does:

``
$ mvn clean package -DskipTests -Dgpg.skip=true
`` 

It does so on do/while blocks, when the "while" statement is put on the same line as the curly bracket.

If this is wanted, then i fixed those violations.

I also added checks for "run.sh" in the jmh folder, so that it fails, whenever one of the "mvn" commands fails. Right now it always ran through and jmh used old deployed versions of the jars (grrrr, took me some time to figure out, that it failed bcs of checkstyle)

### DIFF
--- a/jmh/run.sh
+++ b/jmh/run.sh
@@ -2,18 +2,23 @@
 BASEDIR=$(dirname $0)
 
 
-
 echo "Building RoaringBitmap jar"
 rm -f $BASEDIR/../target/RoaringBitmap*.jar
 mvn -f $BASEDIR/../pom.xml clean install -DskipTests -Dgpg.skip=true
+
+[[ $? -eq 0 ]] || exit
 
 echo "Building Real Roaring Dataset jar"
 rm -f $BASEDIR/../real-roaring-dataset/target/real-roaring-dataset*.jar
 mvn -f $BASEDIR/../real-roaring-dataset/pom.xml clean install
 
+[[ $? -eq 0 ]] || exit
+
 echo "Building benchmarks jar"
 rm -f  $BASEDIR/target/benchmarks.jar
 mvn -f $BASEDIR/pom.xml clean install -Dtest=*$1* -DfailIfNoTests=false
+
+[[ $? -eq 0 ]] || exit
 
 echo "Running benchmarks"
 java -jar $BASEDIR/target/benchmarks.jar  true -wi 5 -i 5 -f 1 $1
--- a/pom.xml
+++ b/pom.xml
@@ -94,7 +94,7 @@
           <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>LATEST</version>
+            <version>6.19</version>
           </dependency>
           <dependency>
             <groupId>org.apache.maven.plugins</groupId>
--- a/style/roaring_google_checks.xml
+++ b/style/roaring_google_checks.xml
@@ -57,7 +57,7 @@
         <module name="RightCurly"/>
         <module name="RightCurly">
             <property name="option" value="alone"/>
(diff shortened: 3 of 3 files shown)


okrische:
Checkstyle also fails, when compiled/run with jdk7, hmm.

``
$ mvn clean checkstyle:checkstyle 
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle (default-cli) on project RoaringBitmap: Execution default-cli of goal org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-checkstyle-plugin:2.17:checkstyle: java.lang.UnsupportedClassVersionError: com/puppycrawl/tools/checkstyle/api/AuditListener : Unsupported major.minor version 52.0
...
``

Looks like, checkstyle already works with jdk8? Sigh.

gssiyankai:
Hi @okrische,
The build is failing because:
- Roaring depends on the latest version of com.puppycrawl.tools:checkstyle
- com.puppycrawl.tools:checkstyle:0.7 has been released 2 weeks ago
- this version is not compatible with java 7

To fix the issue, just set explicitly the version to 6.19 in the root pom.

HTH

okrische:
So, with 7.0 then, the RightCurly check will fail with the current master. Ok. Do you think, the RightCurly changes make sense nonetheless? 

I am pretty used to this:

``
do {
   something();
} while(test());
``

But with 7.0 of the puppy checkstyle thingie, together with google style, it enforces:

``
do {
   something();
} 
while(test());
``

lemire:
@okrische 

I am with you here... this style rule seems a tad odd. Nothing wrong with it per se, but forcing people to write this way is a bit much.

What are your thoughts @gssiyankai ?

okrische:
Maybe the google style guide has not been updated yet on checkstyle side, yet.

- https://google.github.io/styleguide/javaguide.html#s4.1-braces

> Line break after the closing brace, only if that brace terminates a statement or terminates the body of a method, constructor, or named class. For example, there is no line break after the brace if it is followed by else or a comma.

And in a do/while the do {} is followed by "while". Hm.

gssiyankai:
@lemire,
 No preference. I can go with both.

lemire:
@okrische  

I feel strongly that this is correct:

```java
        do {
            System.out.println("Count is: " + count);
            count++;
        } while (count < 11);
```

while this is ugly...

```java
        do {
            System.out.println("Count is: " + count);
            count++;
        } 
        while (count < 11);
```

If someone feels very differently, we can discuss it... but if not...

> Maybe the google style guide has not been updated on checkstyle side, yet.

We do this work actually. Roaring has its own style which is a watered down version of the Google style as interpreted by checkstyle... (meaning, we allow more things) That is, we do not use the default "google style" from checkstyle, instead we provide our own.  This gives us some flexibility.

Can you go to file ``style/roaring_google_checks.xml``? Find ``<module name="RightCurly">``... next look at the content:

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/style/roaring_google_checks.xml#L60

I think that the problem is that we have ``LITERAL_DO`` in the list which forces you to make the changes from your commit...

https://github.com/RoaringBitmap/RoaringBitmap/pull/111/commits/f3617a045e90f4ad59c1722d3233e18f1e7834ee

I think that if you delete LITERAL_DO from the line 60 in ``style/roaring_google_checks.xml``, you'll find that your "ugly" commit is unnecessary.

Can you check?

(Note: I do not mean to imply that you did ugly work. Quite the opposite. You brought forth an important issue.)

okrische:
* changed roaring_google_checks.xml, 
* set version of puppycrawl checkstyle tool to 6.19, 
* (also tested with puppycrawl checkstyle tool 7.0, so, when roaring goes jdk8, then it will work as well)
* and added that fix to run.sh in jmh, so it will fail, when any build fails

lemire:
Looks great to me. Since @gssiyankai  went over some of this... let us merge.

PR #112, opened 2016-07-21
Add missing shebang to memory/run.sh

maciej:
Using /usr/bin/env bash as suggested by the top SO answer:
http://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang


lemire:
Sure... great!

PR #121, opened 2016-08-20
Fix the bug in member implementation of 'flip' (static version is fine)

akazakov:
Also fix tests of static versions of remove/add/flip to check the result of operations.


akazakov:
fixes #122

lemire:
Looks good. Thanks for catching this. Merging.

c.c. @owenkaser

Issue #63, opened 2016-01-12
Fine tune array-array intersection algorithm


lemire:
When two array containers are intersected, we select the right algorithm according to a heuristic which depends (arbitrarily) on a parameter (64). This parameter was chosen more or less randomly a long time ago. It could be that either a larger or, more likely, a smaller value could be preferable in practice. We should test this out:

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/Util.java#L439-L441

lemire:
c.c. @gssiyankai

lemire:
Solved with RoaringBitmap-0.6.28 for the in-memory (regular) Roaring. Could be further improved with MappeableArrayContainer... will open a new issue.

lemire:
See the new issue... https://github.com/RoaringBitmap/RoaringBitmap/issues/128

PR #127, opened 2016-11-21
Fine tune array-array intersection algorithm #63

ppiotrow:
The aim of the commit is to verify optimal value for Util.unsignedIntersect2by2 parameter.
Currently it's is param=64
The commit contains first version of benchmarking the `galloping` and `local` algorithm variants.
For param={"10","15","20","25","30","35", "40", "50","60", "90"} it generates five different examples:
 * small container S
 * big contained B (B.length = param * S.size)
The container values are generated by Uniform Distribution in version1.
The examples are deterministic, random seed is hardcoded.

### DIFF
--- a/jmh/pom.xml
+++ b/jmh/pom.xml
@@ -38,7 +38,12 @@
         <dependency>
             <groupId>org.roaringbitmap</groupId>
             <artifactId>RoaringBitmap</artifactId>
-            <version>[0.5.6-SNAPSHOT,)</version><!-- we use an open interval, hoping that we pick up the last version --> 
+            <version>[0.6.28-SNAPSHOT,)</version><!-- we use an open interval, hoping that we pick up the last version -->
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+            <version>3.6.1</version>
         </dependency>
         <dependency>
             <groupId>com.zaxxer</groupId>
@@ -65,7 +70,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <jmh.version>[1.10.1,)</jmh.version>
+        <jmh.version>1.16</jmh.version>
         <javac.target>1.7</javac.target>
         <uberjar.name>benchmarks</uberjar.name>
     </properties>
--- a/jmh/pom.xml
+++ b/jmh/pom.xml
@@ -56,6 +56,11 @@
             <version>0.1-SNAPSHOT</version>
         </dependency>
         <dependency>
+            <groupId>me.lemire.integercompression</groupId>
+            <artifactId>JavaFastPFOR</artifactId>
+            <version>[0.1,)</version>
+        </dependency>
+        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <version>4.11</version>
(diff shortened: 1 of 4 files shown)


lemire:
@ppiotrow 

Great. Do you have some numbers to share?

Can you check the spelling of ``UtilBechmark.java``?

ppiotrow:
## number of examples `local` algorithm was faster than `galloping` for given param.
wins

## for every experiment comparison how better was `local` than `galloping` [ops/s] 
comparison

ppiotrow:
Let's treat this as Work In Progress. I'm going to generate container content with different distributions (currently uniform). I uploaded this to gather your suggestions as you probably know best what could be worth to check.

lemire:
@ppiotrow 

So your results suggest that setting the parameter to 32 instead of 64 could be beneficial.

ppiotrow:
At least for Uniform distribution this is true. Of course it is worth to increase granularity and check differences for `param = {30,31,32,33,34,35}`.

lemire:
A non-uniform distribution that we sometimes use is this one: 

https://github.com/lemire/JavaFastPFOR/blob/master/src/main/java/me/lemire/integercompression/synth/ClusteredDataGenerator.java

ppiotrow:
Thanks for that generator, I'll try it. 
Also for documentation purposes this is copy of python notebook I used to obtain images from benchmark results https://gist.github.com/ppiotrow/d5f2a2f813a7ca8202480be579ef021b

lemire:
I like your Python script. It is neat and simple.

ppiotrow:
As the benchmarked code is used only in `and` operations between two array containers it makes no sense to benchmark containers larger than 1024. Am I right? Currently in the benchmark there are defined `SMALLEST_ARRAY = 50` and `BIGGEST_ARRAY = 2^16`. I think those values should be `2` and `1024`.

lemire:
Array containers can contain between 1 and 4096 integers (inclusively).

ppiotrow:
Now the benchmark uses both uniform and clustered distributions in following way:
small(uniform)    with big(uniform)
small(clustered) with big(uniform)
small(uniform)    with big(clustered)
small(clustered) with big(clustered)
I also fixed the data so now generated arrays are between 2 and 4096 elements.
Bellow there are updated results of the benchmark. The gist was also updated.

For every experiment comparison how better was local than galloping [ops/s]
zrzut ekranu 2016-12-03 o 12 49 05
Number of examples local algorithm was faster than galloping for given param.
zrzut ekranu 2016-12-03 o 12 49 17
Number of examples local algorithm was faster than galloping for given param (only **uniform** distributions)
zrzut ekranu 2016-12-03 o 12 49 29
Number of examples local algorithm was faster than galloping for given param (only **clustered** distributions)
zrzut ekranu 2016-12-03 o 12 49 39

ppiotrow:
The next steps I suggest are
* analyze the cases for `param=30` where local was few times faster than galloping and vice versa,
* perform benchmark for `param = {25,26.....33,34,35}` as it looks like there is best param value.

ppiotrow:
unknown
all
uniform
clustered
Based on the first picture which shows detailed comparison of throughput of `local` and `galloping` approach it looks like from `param=26` we tend to have better results for galloping. Even if `local` is sometimes better (for param={26,27...} it is rather win ratio around 1-15%. It is meaningless given that `galloping` wins 1-70% and the big differences are quite frequent.

lemire:
@ppiotrow Speaking for myself, if you were to change the default based on these results, I think we could merge it and issue a new release. Given that the current default was ad hoc, chances are good that it would be generally beneficial.

ppiotrow:
Last result with more experiments
unknown
I will set threshold to 25.

lemire:
@ppiotrow  You may want to make your status in Roaring public https://github.com/orgs/RoaringBitmap/people/ppiotrow

lemire:
LGTM

lemire:
Public release RoaringBitmap-0.6.28

PR #130, opened 2016-12-18
Tune MappeableArrayContainer intersection algorithm

ppiotrow:
unknown-4
unknow
unknown-3
unknown-2

Seems like the optimal parameter for Buffer version is around 35-40.
Next step is to generate more examples and focus on 25-60 range.
I wonder why is there clear split between "uniform" preferring local and "clustered" preferring galloping


ppiotrow:
Based on the results I suggest setting threshold to 34. 
detailed
general
uniforms
clustered

lemire:
Excellent!

Issue #103, opened 2016-04-05
Range operations on bitmap containers can be inefficient


lemire:
A lot of times, we call for a full computation of the bitmap container's cardinality even when it is inefficient to do so. This affects methods such as addRange, removeRange, not, and all operations between a bitmap container and a run container.

ppiotrow:
I'm interested in solving this. Could you give one up-to-date reference to the current codebase where recomputing cardinality could be skipped?

lemire:
@ppiotrow 

Low-hanging fruit...

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/BitmapContainer.java#L122-L135

Basically... one could trace the calls to ``computeCardinality``...
https://github.com/RoaringBitmap/RoaringBitmap/search?utf8=%E2%9C%93&q=computeCardinality
and try to optimize them out. In a lot of cases, it is possible to avoid a full recompute of the cardinality.

Of course, part of the difficulty is having a reasonable benchmark to show that the new code path is likely better... another difficulty is to keep the maintainability of the code, to avoid bugs and extra overhead.

ppiotrow:
Already traced `computeCardinality` calls but still I don't see when it is valid to have "unrepaired" containers.

lemire:
@ppiotrow 

Ah.  Let me be more precise.


What ``computeCardinality`` does is scan the *entire* bitset. The idea is that there are instances where it would be more economical to scan only part of the bitset.

Now let us go back to this function ...

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/BitmapContainer.java#L122-L135

Clearly, if you knew the number of ones in the specified range (begin,end), call this ``x``, then you could know that the new cardinality is ``cardinality - x + end - begin``. So maybe computing the cardinality of the bitset in a specific subrange (say the interval from begin to end) could speed this up. This is just an example, there are many other instances where this general idea applies.

Similarly, if we knew how to compute the cardinality of just a range of a bitset, we could greatly speed up this other function...

https://github.com/RoaringBitmap/RoaringBitmap/blob/8432470aca62558da09a8ae2337318a1b2465d1f/src/main/java/org/roaringbitmap/RunContainer.java#L478-L492

and so forth.

We "kind" of do it already, see

[...]

PR #131, opened 2016-12-29
Performance of range operations on bitmap containers #103

ppiotrow:
Hi, I've started writing code to solve this issue. Description of the problem by @lemire is very clear. I was looking at  method `cardinalityInBitmapWordRange` to get some ideas. I've committed simple test that fails. The question is if this is a bug or the API is not clear (to me)?


ppiotrow:
Thanks, new methods will be very useful in solving #103

lemire:
@ppiotrow  Please test them thoroughly. They should be correct, but I did not test them.

PR #135, opened 2017-01-02
Don't create a new container on every contains() call to an immutable…

gpunti:
Each call to ImmutableRoaringBitmap.contains() triggers the creation of three instances ( 1xByteBuffer, 1xShortBuffer and 1xMappeableContainer).  The pressure on the garbage collector is huge causing pauses of 10ms ( at 6k requests per second ). 

Moreover, we're checking one time and another things that won't change ( what type of container, number of runs, the offset of the container).

Because this is an immutable structure it makes sense to have the containers cached.

### DIFF
--- a/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringArray.java
+++ b/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringArray.java
@@ -13,6 +13,8 @@ import java.nio.LongBuffer;
 import java.nio.ShortBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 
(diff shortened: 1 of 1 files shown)


lemire:
My concern is that this will raise substantially the memory usage of our bitmaps. I don't know much a ConcurrentHashMap weights but it probably quite a bit. I agree that caching is safe here, but it not free.


I think we could code a lighter approach with little effort. Maybe. I will have to look at the code again.

@gssiyankai : would you chime in?

gssiyankai:
Caching is usually a good idea to speed-up execution. The price to pay in terms of memory depends on what is cached and I would recommend to make sure it is not too high. To do so, one can use the [memory benchmarks](https://github.com/RoaringBitmap/RoaringBitmap/tree/master/memory). These benchmarks are quite stable from what I remember.
Then I would recommend to make sure that caching has no negative impact on execution time. Indeed, I have faced some situations before where the outcome was different from what was expected with caching. The reason is the larger the memory footprint, the less the space left for other computations and the more the cache misses and memory swapping. One can use the [jmh benchmarks](https://github.com/RoaringBitmap/RoaringBitmap/tree/master/jmh) to verify that caching does not slow everything down.

gpunti:
I just monitored the gc again and the improvement was huge!

Issue #134, opened 2016-12-30
streamline the implementations of iterators to drastically reduce memory allocation


lemire:
Currently, iterators are implemented by instantiating sub-iterators that run over the containers. This approach requires us to create many more temporary objects than necessary.

A much simpler approach is the one taken by the ``CRoaring`` library where no memory allocation is done... 
https://github.com/RoaringBitmap/CRoaring/blob/master/src/roaring.c#L1156

This should reduce the amount of code *and* reduce the pressure on the garbage collector, which should improve performance in real-life systems.

This is related to https://github.com/RoaringBitmap/RoaringBitmap/issues/44 but distinct.

lemire:
See the following commit for an application of this idea to the ``contains`` function: 
https://github.com/RoaringBitmap/RoaringBitmap/commit/a778e33adb9c651c8c39e60b1eba17a8edc4b0a0

ppiotrow:
What is the reason of having `RoaringIntIterator` https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L49  and `Iterator<Integer>` https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1442 ? Performance of auto boxing? Is their logic different in some way? I understand that it is not possible to have `Iterator<int>`. 

Second question about `PeekableIntIterator` https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/PeekableIntIterator.java Why are those methods important in real word use cases?

lemire:
@ppiotrow 

Yes, we expect to get better performance with ``IntIterator`` than ``Iterator<Integer>``. It seems that we do too...

        cd RoaringBitmap
        cd jmh
        mvn -f ../pom.xml clean install -DskipTests -Dgpg.skip=true
        mvn -f ../real-roaring-dataset/pom.xml clean install
        mvn  clean install  -DskipTests
        java -jar ./target/benchmarks.jar true -wi 5 -i 5 -f1 IteratorsBenchmark

        IteratorsBenchmark.testBoxed_a             avgt    5     1132.022 ±   1822.187  us/op
        IteratorsBenchmark.testFlyweight_a         avgt    5      651.920 ±      0.169  us/op
[...]

PR #136, opened 2017-01-03
Don't create a new container on every contains() call to an immutable mapped bitmap

gpunti:
Each call to ImmutableRoaringBitmap.contains() triggers the creation of three instances ( 1xByteBuffer, 1xShortBuffer and 1xMappeableContainer). The pressure on the garbage collector is huge causing pauses of 10ms ( at 6k requests per second ).

Moreover, we're checking one time and another things that won't change ( what type of container, number of runs, the offset of the container).

**In this lighter approach we don't use a Map, but a MappeableContainer[] as it is used in its mutable counterpart MutableRoaringArray.**

### DIFF
--- a/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringArray.java
+++ b/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringArray.java
@@ -13,6 +13,8 @@ import java.nio.LongBuffer;
 import java.nio.ShortBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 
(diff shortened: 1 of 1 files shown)


gpunti:
it yet uses too much memory

lemire:
@gpunti I think the following commit solves the issue...

https://github.com/RoaringBitmap/RoaringBitmap/commit/a778e33adb9c651c8c39e60b1eba17a8edc4b0a0

Can you review the commit please to see whether it does fix your issue?

gpunti:
Hi @lemire 

Yes I think this will fix the issue. Tomorrow I'll check the garbage collection again. It should boost the jmh performance tests as well.

Thank you very much!

lemire:
I should be able to release a new version today.

Note that we should be able to implement the iterators in this manner as well, without allocating temporary objects. See https://github.com/RoaringBitmap/RoaringBitmap/issues/134

c.c. @ppiotrow @gssiyankai

PR #141, opened 2017-01-30
wrong data type for pointer

gpunti:
ImmutableBitmap was trying to reference 65k elements with a short value instead of using an int value ( as MutableBitmap does ).

### DIFF
--- a/src/test/java/org/roaringbitmap/buffer/TestIterators.java
+++ b/src/test/java/org/roaringbitmap/buffer/TestIterators.java
@@ -4,246 +4,273 @@
 
 package org.roaringbitmap.buffer;
 
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
-import com.google.common.primitives.Ints;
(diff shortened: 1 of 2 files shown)


src/test/java/org/roaringbitmap/buffer/TestIterators.java
```diff
@@ -245,5 +250,31 @@ public void testEmptySkips() {
...
     it.advanceIfNeeded(0);
   }
+  
+  @Test
+  public void testIteratorsOnLargeBitmap() throws IOException {
+      MutableRoaringBitmap bitmap = new MutableRoaringBitmap();
+
+      int inc = Short.MAX_VALUE;
+
+      for (long i = -Integer.MIN_VALUE; i < Integer.MAX_VALUE; i += inc) {
```
ppiotrow: @gpunti Fix and test are correct but there is small anomaly hidden with overflow of  `- Integer.MIN_VALUE`, see http://stackoverflow.com/questions/12535095/java-integers-min-value-negative-then-compare

lemire:
Excellent.

lemire:
Released the new code.

Issue #145, opened 2017-02-10
Consider receiving Iterator<? extends RoaringBitmap> in FastAggregation functions


Kineolyan:
Hello everyone,

could we consider replacing `Iterator<RoaringBitmap>` by `Iterator<? extends RoaringBitmap>` in all functions of FastAggregation ?
This is a common pattern used by classes consuming values - `List#addAll` for example.
This is already supported natively by the functions taking Arrays.

Regarding code compatibility, it does not break any API as it is an extension of the parameter constraints.

Best regards

lemire:
Sounds reasonable but we would need to catalog all relevant instances... e.g. https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L654

Do you want to issue a PR?

Kineolyan:
Ok. I will have a look at all the usages of Iterator in the code base and translate it when possible with wildcard.
I post the link to the PR when it is ready.

lemire:
@Kineolyan If you could throw in a unit test to check that the extended usage does what one would expect (even if the test is silly) that'd be great.

lemire:
I think that the PR solves this issue. It will be part of the next release.

PR #146, opened 2017-02-14
Using wildcards

Kineolyan:
Improved API for `RoaringBitmap` and `FastAggregation` using wildcards for Iterator and List, as discussed in [Issue 145](https://github.com/RoaringBitmap/RoaringBitmap/issues/145).

Note that one method was not updated: `RoaringBitmap#and(Iterator)` as it was using old iterators - my guess being for compatibility reasons.

Cheers


lemire:
Looks good.

PR #147, opened 2017-02-16
implementation of subset relations - https://github.com/RoaringBitmap…

openkappa:
…/RoaringBitmap/issues/143

I've ported the C code for the subset relations from CRoaring. I implemented this relation as a contains(RoaringBitmap) on RoaringBitmap, so that if A = {1, 2, 3} and B = {2} A.contains(B) would be true, whereas B.contains(A) would be false.


lemire:
+1

openkappa:
I found a transposition error in the last line of RoaringBitmap::contains my first commit. Is it better to rebase now?

gssiyankai:
It does not really matter. You can leave it as such or rebase if you prefer.

lemire:
@openkappa  LGTM.

PR #148, opened 2017-02-17
138 - implement first and last. https://github.com/RoaringBitmap/Roar…

openkappa:
…ingBitmap/issues/138

Following on from comments on issue page this implementation adds first() and last() to container implementations. The offset issue is resolved within RoaringArray, which is delegated to by RoaringBitmap. Tests added for a range of values for each container implementation and RoaringBitmap itself.

### DIFF
--- a/src/main/java/org/roaringbitmap/RoaringArray.java
+++ b/src/main/java/org/roaringbitmap/RoaringArray.java
@@ -578,4 +578,30 @@ public final class RoaringArray implements Cloneable, Externalizable {
   public void writeExternal(ObjectOutput out) throws IOException {
     serialize(out);
   }
+
+  /**
+   * Gets the first value in the array
+   * @return te first value in the array
+   */
+  public int first() {
+    if(size == 0) {
+      return 0;
+    }
+    short firstKey = keys[0];
+    Container container = values[0];
+    return firstKey << 16 | container.first();
+  }
+
+  /**
+   * Gets the last value in the array
+   * @return te last value in the array
+   */
+  public int last() {
+    if(size == 0) {
+      return 0;
+    }
+    short lastKey = keys[size - 1];
+    Container container = values[size - 1];
+    return lastKey << 16 | container.last();
+  }
 }
--- a/src/main/java/org/roaringbitmap/RoaringArray.java
+++ b/src/main/java/org/roaringbitmap/RoaringArray.java
@@ -7,6 +7,7 @@ package org.roaringbitmap;
 
 import java.io.*;
 import java.util.Arrays;
+import java.util.NoSuchElementException;
 
 
 /**
@@ -581,12 +582,11 @@ public final class RoaringArray implements Cloneable, Externalizable {
 
   /**
    * Gets the first value in the array
-   * @return te first value in the array
+   * @return the first value in the array
+   * @throws NoSuchElementException if empty
    */
   public int first() {
-    if(size == 0) {
-      return 0;
-    }
+    assertNonEmpty();
     short firstKey = keys[0];
     Container container = values[0];
     return firstKey << 16 | container.first();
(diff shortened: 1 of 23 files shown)


lemire:
+1

gssiyankai:
LGTM

lemire:
@gssiyankai  Thanks. 

This needs to be ported to the buffer package before we can produce a release. Not hard. Mostly copy and paste. I can do it later... or maybe someone wants to jump in and do it... ;-) (It is probably a 5-minute job.)

openkappa:
I can do it later. I don't actually use the buffer package - is it effectively a carbon copy of the non persistent implementation?

lemire:
@openkappa 

Yes, it is a carbon copy that works over ByteBuffers instead of arrays. It is annoying that we need to maintain two distinct versions of the same code, but there are significant performance trade-offs involved. The ByteBuffer approach works with off-heap processing... so it can avoid the garbage collection... but it is slower compared to working directly on Java arrays.

Mostly, it is a minor annoyance involving copying and pasting.

openkappa:
Buffer version is much more verbose. The branch coverage is going to decrease with a port unless cases can be contrived where the buffer hasn't got an array. Any thoughts on contriving cases to cover array backed buffers and the converse?

src/main/java/org/roaringbitmap/RoaringArray.java
```diff
@@ -578,4 +578,30 @@ protected int size() {
...
     serialize(out);
   }
+
+  /**
+   * Gets the first value in the array
+   * @return te first value in the array
+   */
+  public int first() {
+    if(size == 0) {
+      return 0;
```
Kineolyan: Is this the good policy to return 0 as the value when there is no elements in the RoaringArray (or a RoaringBitmap) ?
I thought that 0 would be a valid value. We may simply throw an error `NoSuchElementException`, as Collections.max does in the JDK.

src/test/java/org/roaringbitmap/TestRoaringBitmap.java
```diff
@@ -4342,4 +4342,49 @@ public void testLazyOr() {
...
+
+  @Test
+  public void testFirstLast_CreateSparseContainersAfterRun() {
+    RoaringBitmap rb = new RoaringBitmap();
+    rb.add(1L, 1 << 14);
+    for(int i = 18; i < 31; ++i) {
+      int x = 1 << i;
+      rb.add(x);
+      Assert.assertEquals(1, rb.first());
+      rb.last();
```
Kineolyan: Is this call mandatory ? You are not storing the last value and computing it again on the next line.

src/main/java/org/roaringbitmap/RoaringArray.java
openkappa: I agree. I'm not sure if Collections::max relates to this functionality, but SortedSet::first, SortedSet::last throw NoSuchElementException for empty sets. I must admit until now I had only considered bitmaps containing natural numbers, but to be consistent with the comments on the PositiveIterator thread the only option would be to throw.

Kineolyan:
Good job @openkappa

src/main/java/org/roaringbitmap/RoaringArray.java
Kineolyan: Yes, Collections::max was the first that came to my mind.
I thought about 0 as a possible value as I know that when using bitmaps as index for lines of DB tables, those lines are refered by their index as in an array - starting from 0.
Anyway, I like it very much now :)

openkappa:
The last commit incurred a performance penalty for run containers. I've fixed that and added explicit tests for cases where the bitmap has been lazily mutated.

lemire:
LGTM.

PR #150, opened 2017-02-22
Streamline iterators

ppiotrow:


ppiotrow:
I start this PR with benchmark results of current implementation. In next commits I'll replace current implementations with code that supposed to be ported from C. However right now after my experiments it more reminds FlyWeight implementation with some improvements.
```
Benchmark                                  Mode  Cnt        Score       Error  Units
IteratorsBenchmark.testBoxed_a             avgt   15      817.559 ±     2.420  us/op
IteratorsBenchmark.testBoxed_b             avgt   15   170350.078 ±   709.025  us/op
IteratorsBenchmark.testBoxed_c             avgt   15  1577855.057 ±  6118.798  us/op
IteratorsBenchmark.testBoxed_run           avgt   15   267327.538 ±   142.615  us/op
IteratorsBenchmark.testFlyweight_a         avgt   15      654.497 ±     0.326  us/op
IteratorsBenchmark.testFlyweight_b         avgt   15   131437.377 ±    31.528  us/op
IteratorsBenchmark.testFlyweight_c         avgt   15  1390726.455 ± 20872.818  us/op
IteratorsBenchmark.testFlyweight_run       avgt   15   223651.291 ±    11.464  us/op
IteratorsBenchmark.testForeach_a           avgt   15      408.408 ±     5.740  us/op
IteratorsBenchmark.testForeach_b           avgt   15     5151.332 ±    17.118  us/op
IteratorsBenchmark.testForeach_c           avgt   15   353001.497 ± 10284.912  us/op
IteratorsBenchmark.testForeach_run         avgt   15     3123.961 ±    20.505  us/op
IteratorsBenchmark.testReverseFlyweight_a  avgt   15      617.025 ±     0.247  us/op
IteratorsBenchmark.testReverseFlyweight_b  avgt   15   101008.467 ±    26.532  us/op
IteratorsBenchmark.testReverseFlyweight_c  avgt   15  1176880.412 ± 13828.394  us/op
IteratorsBenchmark.testReverseStandard_a   avgt   15      603.022 ±     0.702  us/op
IteratorsBenchmark.testReverseStandard_b   avgt   15   100881.164 ±    42.564  us/op
IteratorsBenchmark.testReverseStandard_c   avgt   15  1133575.788 ±   435.108  us/op
IteratorsBenchmark.testStandard_a          avgt   15      696.223 ±     1.083  us/op
IteratorsBenchmark.testStandard_b          avgt   15   130094.878 ±    13.408  us/op
IteratorsBenchmark.testStandard_c          avgt   15  1486500.911 ± 18610.332  us/op
IteratorsBenchmark.testStandard_run        avgt   15   220238.223 ±   530.738  us/op
```

ppiotrow:
Iterator<Integer> now has performance of Flyweight version in 3 tests, but performance on run containers dropped. Already found the reason and the fix will be ready soon.
```
IteratorsBenchmark.testBoxed_a    avgt   15      648.864 ±    0.116  us/op
IteratorsBenchmark.testBoxed_b    avgt   15   132363.647 ±  532.607  us/op
IteratorsBenchmark.testBoxed_c    avgt   15  1368257.699 ± 5188.556  us/op
IteratorsBenchmark.testBoxed_run  avgt   15   329737.888 ±  342.606  us/op
```

src/main/java/org/roaringbitmap/RoaringIterator.java
```diff
@@ -0,0 +1,138 @@
...
+
+  public boolean hasNext() {
+    return containerIndex <= parent.highLowContainer.size;
+  }
+
+  public int next() {
+    int answer;
+    switch (typecode) {
+      case TypeCode.BITSET_CONTAINER:
+        long t = w & -w;
```
openkappa: Could this not be done polymorphically on Container? It seems to be that the type codes are used in several places in CRoaring (where indeed polymorphism _is_ used in the equivalent java features) because there is just no runtime override resolution in C.

ppiotrow: First it was just port from C, but it seems that by this approach I'm able to call `nextContainer();` without additionally checking already checked conditions. Using polymorphism is what Flyweight implementation already does and it seems that this PR is on the good way to beat performance. Anyway, I suggest to keep this review and try to answer it once again at the end.

lemire: @openkappa I think that @ppiotrow  explains it well. 

The current iterators have polymorphic calls and lots of short-lived temporary objects. Each and every call to ``next`` goes through a lot of code, not to mention the pressure it puts on the garbage collector. We can hope for the JVM to optimize it all out, but I am pessimistic.


It is not always inefficient,  but in some cases, I suspect that it is.

Would replacement code be less elegant? Possibly, but it is not like we need to play with this code a lot. I think we mostly want to it to be fast.

openkappa: Fair enough. So the nature of the optimisation is technical, rather than algorithmic. I wasn't particularly looking to create a review as much as to comment on something that stuck out slightly. I'll close the review unless it's helpful to take a look as the PR progresses.

ppiotrow:
Now I replaced `getIntIterator()`. This improved performance, but I need to add code for `advanceIfNeeded` and `peekNext` which currently causes tests to fail.
```
IteratorsBenchmark.testStandard_a    avgt   15      539.738 ±    0.385  us/op
IteratorsBenchmark.testStandard_b    avgt   15    77887.139 ±  222.226  us/op
IteratorsBenchmark.testStandard_c    avgt   15  1005537.061 ± 1043.818  us/op
IteratorsBenchmark.testStandard_run  avgt   15   116377.703 ± 1056.091  us/op
```
What is interesting the "run" case is now twice as fast as Flyweight on the same bitmap.

ppiotrow:
I need to verify benchmark results tomorrow. It seems that the "run" iteration performance hasn't dropped and some performance improvements aren't that big.

ppiotrow:
I'm giving up and this pull request. The source code will be present in my local branch if someone is willing to experiment. Initial benchmark results were not valid and it seems there is no speed improvement in the new approach. Maybe benchmarking iterations is harder than I expected.

PR #151, opened 2017-03-07
Fix MappeableRunContainer rank computation

maciej:
In rare cases calling .rank() method on a run container returns a value that's off by one.

The provided test case reproduces the issue.
I include a proposed fix.


ppiotrow:
You could fix it also here https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RunContainer.java#L1941

lemire:
Looks good to me. Will issue a new release shortly.

lemire:
Latest release contains the fix.

ppiotrow:
@lemire What about in-memory RunContainer? It has same issue.

lemire:
Good point. I had added a test and thought it was fine, but I had missed the "@test" annotation so that the test did not run. There is indeed a bug. I will issue a new release.

PR #154, opened 2017-03-18
Fix RunContainer.andCardinality(RunContainer)

gssiyankai:
Fixes #153

Issue #149, opened 2017-02-20
xorCardinality and fuzzyEquals methods


openkappa:
I have a use case where I want to know with very low latency whether the edit distance between two bitmaps is small, in which case they should be considered to be fuzzily equal. I would also like to be able to rank bitmaps by their edit distance from a given bitmap.

Currently it is possible to implement these functions using the method andCardinality() using the identities |a xor b| = |a or b| - |a and b| and |a xor b| < threshold, but there are numerous opportunities for optimisation by implementing the methods:

int xorCardinality(RoaringBitmap other);

boolean fuzzyEquals(RoaringBitmap other, int threshold);

Are these methods a welcome addition to the interface?

lemire:
Of course. I think that there is no issue whatsoever with ``xorCardinality``.

For ``fuzzyEquals``, I would invite you to discuss it. I guess you want an API like that for performance issues? So that the computation can return false early without processing all of the data?

openkappa:
Precisely - to return as soon as the threshold on edit distance is breached. Whether that's the kind of relation that was envisaged for this library, I don't know.

I will submit a pull request for xorCardinality this week.

lemire:
> Precisely - to return as soon as the threshold on edit distance is breached. Whether that's the kind of relation that was envisaged for this library, I don't know.

Sure it is. What measure of "fuzzy equality" are you thinking about?

openkappa:
Just the edit distance which is a bitwise xor. If |A xor B| < x then A and B are fuzzily equal. If the parameter x is set to zero then the relation becomes strict equality, which would seem consistent from an API point of view.

It's not much more effort to implement Jaccard distance which has |A or B| as a denominator.

lemire:
> It's not much more effort to implement Jaccard distance which has |A or B| as a denominator.

It is harder to bail out early though, no?

lemire:
My main worry is one of terminology. We want to make sure that the API is intuitive. I am not sure ``fuzzyEqual`` is the best term, though maybe it is.

What you are actually measuring is the Hamming distance. (I think that the edit distance usually refers to strings.) So what about ``hasSmallHammingDistance``?

https://en.wikipedia.org/wiki/Hamming_distance

openkappa:
It could be poor terminology on my part.

[...]

PR #155, opened 2017-03-18
Xor cardinality

openkappa:
This is a pull request in reference to the symmetric difference/xorCardinality part of #149. I added the left difference as andNotCardinality as well to be consistent with andCardinality and orCardinality.

This pull request depends on #154.


lemire:
This one seems like a no-brainer.

@gssiyankai  want to have a look?

gssiyankai:
LGTM

PR #156, opened 2017-03-18
Hamming similarity

openkappa:
This is a pull request in reference to the fuzzy equals/hamming distance part of #149. The method will return early container by container. I have a _slightly_ faster but _significantly_ more complicated implementation to return early when the threshold is breached at the container level that I could propose if we can agree on the method name and whether this function belongs in the library.

This pull request depends on #154.

### DIFF
--- a/src/main/java/org/roaringbitmap/Container.java
+++ b/src/main/java/org/roaringbitmap/Container.java
@@ -122,6 +122,16 @@ public abstract class Container implements Iterable<Short>, Cloneable, Externali
     }
   }
 
+  /**
+   * Returns the cardinality of the XOR between the passed container and this
+   * container without materialising a temporary container.
+   * @param other other container
+   * @return the cardinality of the symmetric difference of the two containers
+   */
+  public int xorCardinality(Container other) {
+    return getCardinality() + other.getCardinality() - 2 * andCardinality(other);
+  }
+
 
   /**
    * Computes the bitwise ANDNOT of this container with another (difference). This container as well
--- a/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -743,6 +743,31 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
     return x1.getCardinality() + x2.getCardinality() - andCardinality(x1, x2);
   }
 
+  /**
+   * Cardinality of the bitwise XOR (symmetric difference) operation.
+   * The provided bitmaps are *not* modified. This operation is thread-safe
+   * as long as the provided bitmaps remain unchanged.
+   *
+   * @param x1 first bitmap
+   * @param x2 other bitmap
+   * @return cardinality of the symmetric difference
+   */
+  public static int xorCardinality(final RoaringBitmap x1, final RoaringBitmap x2) {
+    return x1.getCardinality() + x2.getCardinality() - 2 * andCardinality(x1, x2);
+  }
+
+  /**
+   * Cardinality of the bitwise ANDNOT (left difference) operation.
+   * The provided bitmaps are *not* modified. This operation is thread-safe
+   * as long as the provided bitmaps remain unchanged.
+   *
+   * @param x1 first bitmap
+   * @param x2 other bitmap
+   * @return cardinality of the left difference
+   */
+  public static int andNotCardinality(final RoaringBitmap x1, final RoaringBitmap x2) {
+    return x1.getCardinality() - andCardinality(x1, x2);
+  }
 
   /**
    * Generate a new bitmap with all integers in [rangeStart,rangeEnd) removed.
@@ -1237,6 +1262,49 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
       return srb.highLowContainer.equals(this.highLowContainer);
     }
     return false;
+  }
+
+  /**
+   * Returns true if the other bitmap has no more than tolerance bits
+   * differing from this bitmap. The other may be transformed into a bitmap equal
+   * to this bitmap in no more than tolerance bit flips if this method returns true.
+   *
+   * @param other the bitmap to compare to
+   * @param tolerance the maximum number of bits that may differ
+   * @return true if the number of differing bits is smaller than tolerance
+   */
+  public boolean isHammingSimilar(RoaringBitmap other, int tolerance) {
+    final int size1 = highLowContainer.size();
+    final int size2 = other.highLowContainer.size();
+    int pos1 = 0;
+    int pos2 = 0;
+    int budget = tolerance;
+    while(budget >= 0 && pos1 < size1 && pos2 < size2) {
+      final short key1 = this.highLowContainer.getKeyAtIndex(pos1);
+      final short key2 = other.highLowContainer.getKeyAtIndex(pos2);
+      Container left = highLowContainer.getContainerAtIndex(pos1);
+      Container right = other.highLowContainer.getContainerAtIndex(pos2);
+      if(key1 == key2) {
+        budget -= left.xorCardinality(right);
+        ++pos1;
+        ++pos2;
+      } else if(key1 < key2) {
+        budget -= left.getCardinality();
+        ++pos1;
+      } else {
+        budget -= right.getCardinality();
+        ++pos2;
+      }
+    }
+    while(budget >= 0 && pos1 < size1) {
+      Container container = highLowContainer.getContainerAtIndex(pos1++);
+      budget -= container.getCardinality();
+    }
+    while(budget >= 0 && pos2 < size2) {
+      Container container = other.highLowContainer.getContainerAtIndex(pos2++);
+      budget -= container.getCardinality();
+    }
+    return budget >= 0;
   }
 
   /**
--- a/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -1288,7 +1288,7 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
         budget -= left.xorCardinality(right);
         ++pos1;
         ++pos2;
-      } else if(key1 < key2) {
+      } else if(Util.compareUnsigned(key1, key2) < 0) {
         budget -= left.getCardinality();
         ++pos1;
       } else {
--- a/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
+++ b/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
@@ -4,14 +4,7 @@
 
 package org.roaringbitmap.buffer;
 
-import org.roaringbitmap.ImmutableBitmapDataProvider;
-import org.roaringbitmap.IntConsumer;
-import org.roaringbitmap.IntIterator;
-import org.roaringbitmap.PeekableIntIterator;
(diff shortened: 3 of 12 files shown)


lemire:
Also looks good to me. I could imagine making it faster by going down to the container level but it would be a *lot* more work and would not be always faster. 

@gssiyankai  want to have a look?

gssiyankai:
@openkappa 
Can you resolve conflicts, please ?

src/main/java/org/roaringbitmap/RoaringBitmap.java
```diff
@@ -1265,6 +1265,51 @@ public boolean equals(Object o) {
...
+    int pos1 = 0;
+    int pos2 = 0;
+    int budget = tolerance;
+    while(budget >= 0 && pos1 < size1 && pos2 < size2) {
+      final short key1 = this.highLowContainer.getKeyAtIndex(pos1);
+      final short key2 = other.highLowContainer.getKeyAtIndex(pos2);
+      Container left = highLowContainer.getContainerAtIndex(pos1);
+      Container right = other.highLowContainer.getContainerAtIndex(pos2);
+      if(key1 == key2) {
+        budget -= (left.getCardinality()
```
gssiyankai: Implementing XorCardinality at the container level is an improvement that sounds promising to me...

src/main/java/org/roaringbitmap/RoaringBitmap.java
```diff
@@ -1265,6 +1265,51 @@ public boolean equals(Object o) {
...
+    final int size2 = other.highLowContainer.size();
+    int pos1 = 0;
+    int pos2 = 0;
+    int budget = tolerance;
+    while(budget >= 0 && pos1 < size1 && pos2 < size2) {
+      final short key1 = this.highLowContainer.getKeyAtIndex(pos1);
+      final short key2 = other.highLowContainer.getKeyAtIndex(pos2);
+      Container left = highLowContainer.getContainerAtIndex(pos1);
+      Container right = other.highLowContainer.getContainerAtIndex(pos2);
+      if(key1 == key2) {
```
gssiyankai: What about a branchless algorithm here ? all containers will be traversed anyway, as long as ```budget``` is positive and there is no galloping search to overcome branch misprediction penalty ...

src/main/java/org/roaringbitmap/RoaringBitmap.java
openkappa: Certainly.

src/main/java/org/roaringbitmap/RoaringBitmap.java
openkappa: I don't know if I fully understand. Are you suggesting that if the keys aren't equal, for the sake of argument, key1 < key2 we should advance pos1 until key1 == key2, and decrement budget by left.cardinality for each container? 

I guess the idea behind was that most bitmaps aren't hamming similar, and it's better to get out as soon as the budget for differences is exhausted, hence a position by position by iteration. getCardinality is very cheap though so it might be a bit faster, if I have interpreted the term "gallop" correctly.

openkappa:
I put each suggestion from the review in a separate commit. Let me know if 1057515 was what you meant or not.

src/main/java/org/roaringbitmap/Container.java
```diff
@@ -122,6 +122,16 @@ public int andCardinality(Container x) {
...
   }
 
+  /**
+   * Returns the cardinality of the XOR between the passed container and this
+   * container without materialising a temporary container.
+   * @param other other container
+   * @return the cardinality of the symmetric difference of the two containers
+   */
+  public int xorCardinality(Container other) {
+    return getCardinality() + other.getCardinality() - 2 * andCardinality(other);
```
gssiyankai: My idea was to implement ```xorCardinality``` for each type of container vs another container... but on second thoughts, given that ```getCardinality```  is cheap... except for RunContainer... I think the improvement won't be very big... Let's keep it like this... one can try to implement and benchmark later...

src/main/java/org/roaringbitmap/RoaringBitmap.java
gssiyankai: Galloping search is a search algorithm on a sorted array that works in 2 steps. First, you start from a given index and move forward, doubling the size of the step at each iteration, until the current element is greater than the one you are looking for. Then, you do a binary search between the current and the previous element you have seen.

src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
```diff
@@ -1026,6 +1019,53 @@ public boolean equals(Object o) {
...
+   * @return true if the number of differing bits is smaller than tolerance
+   */
+  public boolean isHammingSimilar(ImmutableRoaringBitmap other, int tolerance) {
+    final int size1 = highLowContainer.size();
+    final int size2 = other.highLowContainer.size();
+    int pos1 = 0;
+    int pos2 = 0;
+    int budget = tolerance;
+    while(budget >= 0) {
+      short key1 = this.highLowContainer.getKeyAtIndex(pos1);
```
gssiyankai: I was thinking about something like this:
```java
public boolean isHammingSimilar(RoaringBitmap other, int tolerance) {
    final int size1 = highLowContainer.size();
    final int size2 = other.highLowContainer.size();
    int pos1 = 0;
    int pos2 = 0;
    int budget = tolerance;
    while(budget >= 0 && pos1 < size1 && pos2 < size2) {
      final int key1 = Util.toIntUnsigned(this.highLowContainer.getKeyAtIndex(pos1));
      final int key2 = Util.toIntUnsigned(other.highLowContainer.getKeyAtIndex(pos2));
      final Container c1 = highLowContainer.getContainerAtIndex(pos1);
      final Container c2 = other.highLowContainer.getContainerAtIndex(pos2);
      budget -= ((Integer.signum(key1 - key2) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1)) * c1.getCardinality();
      budget -= ((Integer.signum(key2 - key1) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1)) * c2.getCardinality();
      budget += ((Integer.signum(key1 - key2) + 1) & 0x1) * 2 * c1.andCardinality(c2);
      pos1 += (Integer.signum(key1 - key2) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1);
      pos2 += (Integer.signum(key2 - key1) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1);
    }
    while(budget >= 0 && pos1 < size1) {
      Container container = highLowContainer.getContainerAtIndex(pos1++);
      budget -= container.getCardinality();
    }
    while(budget >= 0 && pos2 < size2) {
      Container container = other.highLowContainer.getContainerAtIndex(pos2++);
      budget -= container.getCardinality();
    }
    return budget >= 0;
  }
```
I'm not sure if it will speed up computations... maybe it would be safer to benchmark with real dataset before using my method...

src/main/java/org/roaringbitmap/Container.java
openkappa: Actually this is how I started off quite some time ago with the XOR cardinality pull request. The code is very complicated consisting of 6 separate container level XOR algorithms, mandating a few code duplications and is probably full of the kind of bugs you find 1 year after release. The venn diagram identity has the benefit of being obviously correct, so long as you can get andCardinality right, without much overhead as you note.

src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java
openkappa: I definitely misunderstood what you were getting at in your original comment, and actually prefer the original implementation to 1057515. Your implementation is sophisticated and I wouldn't have thought to implement the logic that way. I would prefer that either you commit 

`while(budget >= 0 && pos1 < size1 && pos2 < size2) {
      final int key1 = Util.toIntUnsigned(this.highLowContainer.getKeyAtIndex(pos1));
      final int key2 = Util.toIntUnsigned(other.highLowContainer.getKeyAtIndex(pos2));
      final Container c1 = highLowContainer.getContainerAtIndex(pos1);
      final Container c2 = other.highLowContainer.getContainerAtIndex(pos2);
      budget -= ((Integer.signum(key1 - key2) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1)) * c1.getCardinality();
      budget -= ((Integer.signum(key2 - key1) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1)) * c2.getCardinality();
      budget += ((Integer.signum(key1 - key2) + 1) & 0x1) * 2 * c1.andCardinality(c2);
      pos1 += (Integer.signum(key1 - key2) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1);
      pos2 += (Integer.signum(key2 - key1) >>> 31) + ((Integer.signum(key1 - key2) + 1) & 0x1);
    }`

or we revert to the original implementation and back out 1057515. 

The question is whether you'd prefer to be slightly faster to verify that a bitmap is _not_ hamming similar, and avoid calling andCardinality whenever possible, at the expense of being branchy, or if you'd prefer to verify that a bitmap _is_ hamming similar as fast as possible, at the expense of being slower to rule bitmaps out by calling andCardinality more often than necessary.  Being selfish, I need the second use case, and your implementation may well speed up my code that I want to utilise this method from. A benchmark would definitely help decide.

gssiyankai: Let's rollback to the first version. We can keep this as a future possible improvement.

openkappa:
@gssiyankai I have cleaned up the history somewhat and reverted to the original algorithm. As discussed, performance can be tuned later if it becomes an issue.

src/test/java/org/roaringbitmap/buffer/TestContainer.java
```diff
@@ -102,27 +105,27 @@ public void inotTest10() {
     assertTrue(c1 instanceof MappeableArrayContainer);
     assertEquals(14, c1.getCardinality());
     assertTrue(checkContent(c1,
-        new short[] {0, 2, 4, (short) 65190, (short) 65191, (short) 65192, (short) 65193,
-            (short) 65194, (short) 65195, (short) 65196, (short) 65197, (short) 65198,
-            (short) 65199, (short) 65200}));
+            new short[]{0, 2, 4, (short) 65190, (short) 65191, (short) 65192, (short) 65193,
```
gssiyankai: This change is also an accidental reformatting I guess...

gssiyankai: @lemire defined a formatting policy here https://github.com/RoaringBitmap/RoaringBitmap/tree/master/style. It's based on google check style.
HTH

openkappa: Sorry I got into a complete mess on a new machine with no sane defaults set up. I have reverted all unintentional formatting errors - could these last two commits be squashed if the PR is merged?

gssiyankai:
Yes, absolutely.

gssiyankai:
@openkappa 
are you done ? If so, we did not rollback to the first version that consisted in 3 consecutive loops without any inner loops...

openkappa:
As of 095f32c there are no inner loops and we are using the first version of the algorithm. I apologise for the level of churn since "cleaning up" :)

gssiyankai:
Good.

lemire:
@gssiyankai I guess we need to issue a new version of the library?

gssiyankai:
Yes, please.

lemire:
Release 0.6.41

Issue #158, opened 2017-04-14
rank of RoaringBitmap produces incorrect result


MiSawa:
It looks like [this](https://github.com/RoaringBitmap/RoaringBitmap/blob/master/src/main/java/org/roaringbitmap/RoaringBitmap.java#L1780) line should be `else if (Util.compareUnsigned(key, xhigh) == 0)`.
By this line,
```java
RoaringBitmap.bitmapOf(65537).rank(1)
```
returns not `0` but `1`.

gssiyankai:
@MiSawa thanks for reporting the issue. Fixed with #159

gssiyankai:
@lemire can you issue a new release, please ?

lemire:
@gssiyankai  Ongoing.

lemire:
Release done. Should appear on maven repo soon.

PR #159, opened 2017-04-14
rank of RoaringBitmap produces incorrect result #158

openkappa:
The issue is addressed in the buffer package too.


gssiyankai:
@openkappa thanks

Issue #161, opened 2017-05-03
implement fast run-array "equal" comparison


lemire:
See https://github.com/RoaringBitmap/CRoaring/issues/99

gssiyankai:
Done with #162

PR #162, opened 2017-05-05
faster equals comparison run vs array #161

richardstartin:
There's no observable speed up in average speed, but the high centiles are improved significantly (on my laptop).

### DIFF
--- a/src/main/java/org/roaringbitmap/RunContainer.java
+++ b/src/main/java/org/roaringbitmap/RunContainer.java
@@ -948,13 +948,13 @@ public final class RunContainer extends Container implements Cloneable {
   }
 
   private boolean equals(ArrayContainer arrayContainer) {
-    if (getCardinality() != arrayContainer.getCardinality()) {
-      return false;
-    }
     int pos = 0;
     for (short i = 0; i < nbrruns; ++i) {
       short runStart = getValue(i);
       short length = getLength(i);
+      if (pos + length >= arrayContainer.getCardinality()) {
+        return false;
+      }
       if (arrayContainer.content[pos] != runStart) {
         return false;
       }
@@ -963,7 +963,7 @@ public final class RunContainer extends Container implements Cloneable {
       }
       pos += length + 1;
     }
-    return pos == arrayContainer.getCardinality();
+    return true;
   }
 
   @Override
--- a/src/main/java/org/roaringbitmap/RunContainer.java
+++ b/src/main/java/org/roaringbitmap/RunContainer.java
@@ -912,19 +912,9 @@ public final class RunContainer extends Container implements Cloneable {
   @Override
   public boolean equals(Object o) {
     if (o instanceof RunContainer) {
-      RunContainer srb = (RunContainer) o;
-      if (srb.nbrruns != this.nbrruns) {
-        return false;
-      }
-      for (int i = 0; i < nbrruns; ++i) {
-        if (this.getValue(i) != srb.getValue(i)) {
-          return false;
-        }
-        if (this.getLength(i) != srb.getLength(i)) {
-          return false;
-        }
-      }
-      return true;
+      return equals((RunContainer) o);
+    } else if (o instanceof ArrayContainer) {
+      return equals((ArrayContainer) o);
     } else if (o instanceof Container) {
       if (((Container) o).getCardinality() != this.getCardinality()) {
         return false; // should be a frequent branch if they differ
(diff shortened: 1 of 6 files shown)


src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -942,6 +932,40 @@ public boolean equals(Object o) {
...
+      short length = getLength(i);
+      if (arrayContainer.content[pos] != runStart) {
+        return false;
+      }
+      if (arrayContainer.content[pos + length] != runStart + length) {
+        return false;
+      }
+      pos += length + 1;
+    }
+    return pos == arrayContainer.getCardinality();
```
gssiyankai: Why this line ? Can you give me a case where this check is required ?

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -942,6 +932,40 @@ public boolean equals(Object o) {
...
+      }
+      if (getLength(i) != runContainer.getLength(i)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private boolean equals(ArrayContainer arrayContainer) {
+    if (getCardinality() != arrayContainer.getCardinality()) {
```
gssiyankai: Given that we need to traverse the RunContainer to compute its cardinality, this might have a higher cost than checking the cardinality in the loop below, I guess...

src/main/java/org/roaringbitmap/RunContainer.java
richardstartin: Algorithmically, no. I thought it was a nice way of expressing/reinforcing the loop invariant. We can just return true if you prefer...

src/main/java/org/roaringbitmap/RunContainer.java
richardstartin: I'll see what the benchmark says

richardstartin: I ran a few scenarios, the difference in execution speed is hardly noticeable, but (as you'd expect) you can start to see it when you create lots of short runs. _Probabalistically_ the more short runs you have in the range [0,4096) the less likely you are to have an array container with exactly the same bits set, so I doubt you would ever get past this first check when the number of runs is large enough to make a difference. In any case, I changed it, and added a benchmark for many small runs with cardinality less than 4096.

lemire:
@gssiyankai  I think we can issue a release based on this at some point in the future. Let me know.

src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -942,6 +932,40 @@ public boolean equals(Object o) {
...
+    return true;
+  }
+
+  private boolean equals(ArrayContainer arrayContainer) {
+    int pos = 0;
+    for (short i = 0; i < nbrruns; ++i) {
+      short runStart = getValue(i);
+      short length = getLength(i);
+      if (pos + length >= arrayContainer.getCardinality()) {
+        return false;
```
gssiyankai: Can you add unit tests to cover this case ? The code is straightforward but it would nice if we can cover it properly for maintenance concerns.

richardstartin: Sure. In fact, having relied on a broken equals causing havoc throughout the test suite, this code was broken when I removed the method exit condition, which didn't break the rest of the test suite. I've covered all branches now, but it might be worth having some randomised testing on RoaringBitmap.equals because I'm surprised this didn't break the entire suite this morning.

gssiyankai: Indeed, the final cardinality check is mandatory now that the initial one was moved in the loop.
Randomised testing is one way to improve your code coverage and the overall quality of your software. In my opinion, it is useful to spot hidden bugs. But I personally prefer to write unit tests first. I've always seen the return on investment worth it.
Note that if the test suite was not failing this morning, it's just because it was not covering all the new possible cases introduced in the PR. And there was no need to cover them since the existing code was handling all cases in a generic way.

gssiyankai:
Good job! I'll merge the PR soon.

src/main/java/org/roaringbitmap/RunContainer.java
richardstartin: Well, a valid interpretation of what happened is that none of the "shapes" of containers which could break the code as of this morning appear in the test suite anywhere, so you don't actually _know_ they weren't broken before... _Which is not to say that the old code did not handle these "shapes" of containers correctly_. With randomised testing, you'd stand a good chance of producing these failure cases without a priori reasoning, at which point you can crystalise your failure case as a unit test.

gssiyankai: I get your point. It's always interesting to get more insights on testing.
So, do you believe that one should add randomised tests on every feature that is implemented ?

lemire: @richardstartin  The Go counterpart of this library has fuzzy testing: 

https://github.com/RoaringBitmap/roaring/blob/master/Makefile#L71

It is great to have as supplementary testing.

I don't think it replaces or even competes with unit testing.

richardstartin: In a library which models sets of integers, there are dozens of simple invariants (the equals method should be reflexive, if bitmaps are equal, cardinalities should be equal, A and not A and not B should equals A, the list goes on...). Basically there are plenty of hypotheses from which to generate test cases faster than you could possibly think of them...

gssiyankai:
@lemire let's issue a new release

lemire:
Ok. Proceeding.

PR #165, opened 2017-05-22
minor code improvements

ppiotrow:


src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -2275,27 +2275,7 @@ private Container toEfficientContainer() {
...
       return this;
     }
-    if (card <= ArrayContainer.DEFAULT_MAX_SIZE) {
-      ArrayContainer answer = new ArrayContainer(card);
-      answer.cardinality = 0;
-      for (int rlepos = 0; rlepos < this.nbrruns; ++rlepos) {
-        int runStart = Util.toIntUnsigned(this.getValue(rlepos));
-        int runEnd = runStart + Util.toIntUnsigned(this.getLength(rlepos));
-
-        for (int runValue = runStart; runValue <= runEnd; ++runValue) {
```
ppiotrow: It's not possible to apply exactly same change for the Mapped version as there is a branch `BufferUtil.isBackedBySimpleArray` which is valid only for one case.

lemire:
Looks good. Anyone wants to have a look before we merge?

gssiyankai:
LGTM

PR #168, opened 2017-06-07
Negative values documented

ppiotrow:
I was trying to confirm what is expected behaviour of bitmaps with negatives values. 
From the tests it was obvious that arguments are teated as UNSIGNED_INT.
Later I found it nicely documented in README but it would be great to have it documented it in javadoc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Please see the examples folder for more examples.
 Unsigned integers
 ------------------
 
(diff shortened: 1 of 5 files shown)


lemire:
Ok, I will merge this, thanks. 

Note however that the lack of unsigned integers is not due to a limitation of the JVM but rather a limitation of Java itself. So I will make some minor corrections.

ppiotrow:
You're right, it's Java.

PR #175, opened 2017-08-07
Handle signedness correctly in RunContainer.first

gorset:
This fixes a bug where the first value would be returned as a negative
number when the start of the run used all 16 bits.

### DIFF
--- a/src/main/java/org/roaringbitmap/RunContainer.java
+++ b/src/main/java/org/roaringbitmap/RunContainer.java
@@ -2467,7 +2467,7 @@ public final class RunContainer extends Container implements Cloneable {
   @Override
   public int first() {
     assertNonEmpty(numberOfRuns() == 0);
-    return valueslength[0];
+    return Util.toIntUnsigned(valueslength[0]);
   }
 
   @Override
--- a/src/test/java/org/roaringbitmap/TestRunContainer.java
+++ b/src/test/java/org/roaringbitmap/TestRunContainer.java
@@ -3099,6 +3099,13 @@ public class TestRunContainer {
   }
 
   @Test
+  public void testFirstUnsigned() {
+    RoaringBitmap roaringWithRun = new RoaringBitmap();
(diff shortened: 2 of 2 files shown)


richardstartin:
Unless I'm missing something the previous behaviour was correct since Roaring stores unsigned integers. Once you get to 16 bits in the key, you're into the negative signed integers.

@lemire do you agree?

richardstartin:
Sorry, I understand having looked at the new test case. This bug happened whenever a run started after the 16th bit of the container. It doesn't relate to the key at all. The fix looks good to me.

lemire:
Thanks, I am looking at this soon. I'm back from vacations.

gorset:
Thanks for looking at it. We discovered this in production when the following code failed:

```
            RoaringBitmap positions = entry.getValue();
            while (!positions.isEmpty()) {
                RoaringBitmap batch = positions.limit(batchSize);
                positions.remove(batch.first(), batch.last() + 1L);
```
(let me know if there's a more idiomatic way to iterate through a roaring bitmap in batches).

Our workaround was to change `batch.first()` to `batch.select(0)`.

richardstartin:
@gorset you have another production error on its way: RoaringBitmap.last returns negative values when there are 16 bits in the key (its contents are interpreted as unsigned integers). This is by design. 

As far as iteration in batches is concerned, I prototyped cursor based pagination similar to Solr on a branch but didn't take it any further. I could finish that code this month if it's useful.

gorset:
@richardstartin, that's ok. By design we don't have that many possible values. We partition data to make sure we never need more than 31 bits in a single partition.

lemire:
@gorset If you need to page through the values, why don't you do something like this...

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/examples/PagedIterator.java#L14-L21

lemire:
I'll issue a new release. Thanks to all involved.

Issue #109, opened 2016-05-12
Implement support for 64-bit integers


lemire:
We have several implementations of Roaring over 64-bit integers in separate repositories...

https://bitbucket.org/samytto/lazyroaring64-bits

https://bitbucket.org/samytto/roaring64bits2levels

https://bitbucket.org/samytto/roaringtreemap

... as well as an evaluation...

https://bitbucket.org/samytto/datastructures64-bitints

Work is needed to bring the best of these implementations in the main RoaringBitmap library.

luoguohao:
thx,that`s what i want.

JasonHuangHuster:
hi , lemire, i 've checked the "Roaring over 64-bit" code  , it support long quite good , but it's didn't realize any Serialization and Deserialization which eagerly needed in my project .. can use suggest any other java BitMap which support 64 bit ?     BestWishes ! :smiley:

lemire:
@JasonHuangHuster 

I do not. Sorry.

However, you can fork the code and build up the features you need.

ponkin:
+1 for this feature in RoaringBitmap. Now we have to create RoaringBitmap[] to hold huge BloomFilters( up to 10 billions bits)

lemire:
Warning: this issue is marked as "help wanted" meaning that nobody is working on it.

openkappa:
@lemire what about using a long to store the most significant bits in RoaringArray? 

Using a long costs 48 bits per key on the short required for 32 bit storage, and allows storage of up to 80 bit integers. A cost of 48 bits per key is competitive with any solution requiring an object reference (32 bits on a small heap, 64 bits > 64GB heap). Thoughts?

Does an admissible solution require that RoaringBitmaps storing only 32 bit integers do not get larger or slower under any circumstances?

lemire:
@openkappa Yes, what you describe is one possibility out of a few that are quite reasonable. We have a paper http://r-libre.teluq.ca/930/ that reviews these possibilities and there are some experiments. Sadly the paper is in French. There is more work to be done on this front.

openkappa:
[...]

PR #176, opened 2017-08-12
Introduce RoaringTreeMap to handle a bitmap holding any long value

blasd:
Hello,

This refers to https://github.com/RoaringBitmap/RoaringBitmap/issues/109

I needed a bitmap holding any long values to improve management of object references in a fork of MAT. Here is a my current state-of-work: it handles .rank, .select and .getCardinality . I added a few unit-tests.

If this would actually be integrated in RoaringBitmap, please provide feedbacks about what would be the relevant next-steps, especially regarding naming conventions, or API (e.g. should .getCardinality returns a long).


lemire:
That's fantastic and it does feel like something we ought to merge. There is much demand for this feature.

Note that you currently have hundreds of style violations. We enforce the "Google" code style. That's annoying, but the alternative is that people keep on reformatting other people's code. To avoid such bike shedding, we settled on one reasonable format and we try to enforce it at commit time. Assuming that you are using a reasonable IDE, you ought to be able to reformat the code to substantially lower the number of style violations. Would you give it a try? If you can't manage to do it, I can have a look later.

Obviously, we are going to have to do some code review. Ping to @gssiyankai, are you available?

richardstartin:
Hi Benoit, it's been a while. :)

I noticed something related to storing large integers - RoaringBitmap interprets integers as unsigned, and the projects in other languages which have implemented support for longs support all 64 bits. Since a NavigableTreeMap has been used, your implementation supports negative numbers but considers them to precede zero.

Should these tests pass?

 ```
       @Test
	public void testLargeSelectLong() {
		long small = 1;
		long large = 1L << 63 | 1;
		RoaringTreeMap map = new RoaringTreeMap();
		map.addLong(small);
		map.addLong(large);
		long first = map.select(0);
		long last = map.select(1);
		Assert.assertEquals(small, first);
		Assert.assertEquals(large, last);
	}
 ```
 ```
	@Test
	public void testLargeRankLong() {
		long small = 1;
		long large = 1L << 63 | 1;
		RoaringTreeMap map = new RoaringTreeMap();
		map.addLong(small);
		map.addLong(large);
		Assert.assertEquals(2, map.rankLong(large));
	}
 ```
 ```
	@Test
	public void testIterationOrder() {
		long small = 1;
		long large = 1L << 63 | 1;
		RoaringTreeMap map = new RoaringTreeMap();
		map.addLong(small);
		map.addLong(large);
		LongIterator it = map.iterator();
		long first = it.next();
		long last = it.next();
		Assert.assertEquals(small, first);
		Assert.assertEquals(large, last);
	}
```

Richard

**Comment by lemire** [comment]
I think it might be fine to consider longs as signed. Of course, one should document it.

@richardstatin Do you want to volunteer to do a code review of this PR.

**Comment by richardstartin** [comment]
I think the problem with letting RoaringTreeMap be signed is that, while it's the perfect data structure for Benoit's set membership oriented use case in MAT (which has many more applications), and lots of people want a 64 bit structure, it's not the only way of doing 64 bit integers. 

RoaringTreeMap is the wrong data structure for representing ranges: e.g. materialising a full bitmap would take hundreds of GB (see https://github.com/RoaringBitmap/CRoaring/pull/75).  In fact, for use cases other than set membership I think bit slicing, where an unsigned interpretation is preferable, is far superior. So I think it needs to be consistent with better data structures for different use cases.

The code looks good but I feel strongly that the implementation should be unsigned.

File: src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java [hunk 1]
```diff
@@ -0,0 +1,284 @@
...
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import org.roaringbitmap.IntIterator;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+//  this class is not thread-safe
+//@Beta
+public class RoaringTreeMap {
+	protected final NavigableMap<Integer, MutableRoaringBitmap> hiToBitmap = new TreeMap<>();
```
richardstartin: If this is using implementations from the buffer package it should be placed under the buffer package. Otherwise this should be a RoaringBitmap as opposed to MutableRoaringBitmap.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+
+	// Prevent recomputing all cardinalities when requesting consecutive ranks
+	protected boolean allValid = false;
+	protected int firstHiNotValid = Integer.MIN_VALUE;
+
+	// TODO: I would prefer not managing arrays myself
+	protected long[] sortedCumulatedCardinality = new long[0];
+	protected int[] sortedHighs = new int[0];
+
+	protected final List<MutableRoaringBitmap> linkedBitmaps = new ArrayList<>();
```
richardstartin: this is quite similar to the mechanism in RoaringArray where there is an array of short keys and Containers - an improvement could be to factor this state management into something similar to RoaringArray.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+//@Beta
+public class RoaringTreeMap {
+	protected final NavigableMap<Integer, MutableRoaringBitmap> hiToBitmap = new TreeMap<>();
+
+	// Prevent recomputing all cardinalities when requesting consecutive ranks
+	protected boolean allValid = false;
+	protected int firstHiNotValid = Integer.MIN_VALUE;
+
+	// TODO: I would prefer not managing arrays myself
+	protected long[] sortedCumulatedCardinality = new long[0];
```
richardstartin: I worry that this could get quite large, and keeping it synchronised elsewhere in the class is complicated. I would prefer to compute cardinality on the fly and remove this variable.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+			}
+			bitmap.add(y);
+			latest = new AbstractMap.SimpleImmutableEntry<>(x, bitmap);
+		}
+
+		// The cardinalities after this bucket may not be valid anymore
+		firstHiNotValid = Math.min(firstHiNotValid, x);
+	}
+
+	private long pack(int x, int y) {
```
richardstartin: Very minor: could this be moved into Util? As this class gets more attention this operation will be used more and more.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+		if (hiToBitmap.isEmpty()) {
+			return 0L;
+		}
+
+		ensureCumulatives(Integer.MAX_VALUE);
+
+		return sortedCumulatedCardinality[sortedCumulatedCardinality.length - 1];
+	}
+
+	public long select(final long j) {
```
richardstartin: There is a more efficient algorithm in CRoaring which does not require storing the cumulative cardinalities:

    bool select(uint64_t rank, uint64_t *element) const {
        for (const auto& map_entry : roarings) {
            uint64_t sub_cardinality = (uint64_t)map_entry.second.cardinality();
            if (rank < sub_cardinality) {
                *element = ((uint64_t)map_entry.first) << 32;
                // assuming little endian
                return map_entry.second.select((uint32_t)rank, ((uint32_t*)element));
            }
            rank -= sub_cardinality;
        }
        return false;
    }

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+			} else {
+				// The rank is the cardinality of this previous bitmap
+				return sortedCumulatedCardinality[insertionPoint - 1];
+			}
+		}
+	}
+
+	protected void ensureCumulatives(int x) {
+		// Check if missing data to handle this rank
+		if (!allValid && firstHiNotValid <= x) {
```
gssiyankai: Do you assume that `addLong(long)` has to be called with values in ascending order ?
If not, I can think of an edge case where relying on the boolean `allValid` is wrong. The following test is failing and it should not if I am not mistaking:
```scala
@Test
public void test() {
	RoaringTreeMap map = new RoaringTreeMap();
	map.addLong(Long.MAX_VALUE);
	Assert.assertEquals(Long.MAX_VALUE, map.select(0));
	map.addLong(666);
	Assert.assertEquals(666, map.select(0));
	Assert.assertEquals(Long.MAX_VALUE, map.select(0));
}
```

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+//  this class is not thread-safe
+//@Beta
+public class RoaringTreeMap {
+	protected final NavigableMap<Integer, MutableRoaringBitmap> hiToBitmap = new TreeMap<>();
+
+	// Prevent recomputing all cardinalities when requesting consecutive ranks
+	protected boolean allValid = false;
+	protected int firstHiNotValid = Integer.MIN_VALUE;
```
Kineolyan: Consider naming this `firstHighNotValid` for consistency

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+			if (bitmap == null) {
+				bitmap = new MutableRoaringBitmap();
+				hiToBitmap.put(x, bitmap);
+			}
+			bitmap.add(y);
+			latest = new AbstractMap.SimpleImmutableEntry<>(x, bitmap);
+		}
+
+		// The cardinalities after this bucket may not be valid anymore
+		firstHiNotValid = Math.min(firstHiNotValid, x);
```
Kineolyan: A quick optimization here could be to add +1 if x is the latest of sortedHighs.

richardstartin:
What concerns me most is storing the cumulative cardinalities as long[]. It turns out that computing cardinalities on the fly is one of Roaring's strong points. I think how large this structure will be at 10M, 100M, 1BN values comes before access modifiers and variable names in the pertinent hierarchy of needs.

blasd:
Cumulated cardinalities was introduced to prevent the need of iterating through the Map as the iteration cost was unacceptable for my needs. This array has same size as the NavigableMap: adding a long in a long[] seems neglectable to me compared to the price of the Node in the NavigableMap and the weight of an empty Bitmap in the worst case.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I kept the code in a dedicated package to facilitate readibility (at least mine for now :D). But it opens the questions of being able to inject a custom Roaring bitmap implementation, and probable a segregation between an Immutable version and a mutable one

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I need this else performance become too bad for my cases. The memory weight is anyway negligible to the other data-structures (the lenght of the long[] equals the size of the NavigableMap). Moreover, it this grows big, it means each Bitmap holds a small number of ints: this case is one were the TreeRoaringMap will be very weak.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Looks interesting. Could you point out exactly to which code you refers?

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Yes. Moreover, we may need another packing strategy to handle the 64bits behavior. This may be already present in other packing strategies I've coded somewhere else.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Again, the performance hit comes from the NavigableMap iteration. I would need to provide more figures, my original implementation relies on FastUtil, I quickly ported here a version not dependning on FastUtil. TreeMap iteration may be faster but I doubt about this.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+import org.roaringbitmap.IntIterator;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+//  this class is not thread-safe
+//@Beta
+public class RoaringTreeMap {
+	protected final NavigableMap<Integer, MutableRoaringBitmap> hiToBitmap = new TreeMap<>();
+
+	// Prevent recomputing all cardinalities when requesting consecutive ranks
+	protected boolean allValid = false;
```
blasd: My own style-code relevant in other projects. I need to go through the RoaringBitmap official style-sheel.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+	protected int firstHiNotValid = Integer.MIN_VALUE;
+
+	// TODO: I would prefer not managing arrays myself
+	protected long[] sortedCumulatedCardinality = new long[0];
+	protected int[] sortedHighs = new int[0];
+
+	protected final List<MutableRoaringBitmap> linkedBitmaps = new ArrayList<>();
+
+	// Prevent indirection when writing consecutive Integers
+	private transient Map.Entry<Integer, MutableRoaringBitmap> latest = null;
```
blasd: 'latest' is used to prevent the map iteration when adding many consecutive longs. I put transient to indicate this field holds no actual information, not to actual enable Serialization. This has to be done, and then you are right, linkedBitmaps and even cumulatedCardianatis and sortedHighs are not needed for Serialization.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I do not assume this. The idea of ensureCumulatives is to compute it lazily, and dismiss these values on any call to .add/.addLong. For now, any call to .add leads to the dismissing of any ensureCumulatives about the added long.

@lemire, would it be possible/relevant to return a boolean on RoaringBitmap.add, to indicate the Bitmap has been modified (like Collection.add)? It would help refining the dismissing strategy of ensureCumulatives

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+import java.util.NavigableMap;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+import org.roaringbitmap.IntIterator;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+//  this class is not thread-safe
+//@Beta
+public class RoaringTreeMap {
```
blasd: Yes, I need to comply with Roaring style.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I feel you are right. Could you help me by providing a unit-test? Thanks

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+
+					sortedCumulatedCardinality[insertionPosition] =
+							previousCardinality + e.getValue().getLongCardinality();
+
+					if (currentHigh == Integer.MAX_VALUE) {
+						allValid = true;
+						firstHiNotValid = currentHigh;
+					} else {
+						firstHiNotValid = currentHigh + 1;
+					}
```
blasd: I feel you are right. Could you provide a unit-test? Thanks

richardstartin:
@blasd Interesting. Do you have a benchmark to illustrate the marginal benefit in speed and marginal increase in memory consumption?

blasd:
@richardstartin  testPerfManyDifferentBuckets introduces 100.000 different buckets. I've not checked it with TreeMap, but with FastUtil SortedMap, the performance was bad when iterating through all previous bitmaps. Thinking more about it, the performance issue was maybe due to GC: each .reset would instantiate a new iterator. It may be possible to skip the iterator instantiation, and the use of cumulatedCardinalities by using only the sorted list of bitmaps.

richardstartin:
@blasd There is a JMH module in the project for validating decisions like that accurately.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
richardstartin: To reiterate: why are you using the the byte buffer implementation? You will see significantly better performance with RoaringBitmap as opposed to buffer.* classes. If you need to use buffer.* classes, you should move all of longlong underneath org.roaringbitmap.buffer.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
richardstartin: I would be interested to see a benchmark, and what the performance would be if you used RoaringBitmap rather than MutableRoaringBitmap. I would expect the cost of iteration to be minimal compared to the time spent inside Containers.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
Kineolyan: Ok. Then, you should make all performance attributes transient as well, as they duplicate information. I am thinking about allValid, firstHiNotValid, sorted* and linkedBitmaps. LinkedBitmaps and sortedHighs can be rebuilt during deserialization, the others are already handled by a lazy mechanism to be populated on demand.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
Kineolyan: Sure, as soon as I am in front of a computer with Internet, instead of my phone ☺

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Anyway the test fails. I'll have a deeper look. Could you refine the last 2 lines: should map.select(0) returns 666 or Long.MAX_VALUE?

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I do no follow you: if the bitmap is empty, and we write in bucket (variable x) N, then N is the latest of sortedHighs and firstHiNotValid should be N, nor N+1. Thanks for elaborating

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+
+	public long select(final long j) {
+		ensureCumulatives(Integer.MAX_VALUE);
+
+		int position = Arrays.binarySearch(sortedCumulatedCardinality, 0, sortedCumulatedCardinality.length, j);
+
+		if (position >= 0) {
+			// There is a bucket leading to this cardinality: the j-th element is the first element of next bucket
+			MutableRoaringBitmap nextBitmap = linkedBitmaps.get(position + 1);
+			return pack(sortedHighs[position + 1], nextBitmap.first());
```
blasd: Reproduced by:

    @Test
	public void testAddOneSelect2() {
		RoaringTreeMap map = new RoaringTreeMap();

		map.addLong(123);

		Assert.assertEquals(123, map.select(1));
	}

Thanks

blasd: More details: if j is the size of the bitmap, position matches the last cumulated cardinality while selecting this cardinality should returns an 

    throw new IllegalArgumentException(
        "select " + j + " when the cardinality is " + this.getCardinality());

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
gssiyankai: Sorry about that. map.select(0) should return 666 and map.select(1) should return Long.MAX_VALUE

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+
+	protected void ensureCumulatives(int x) {
+		// Check if missing data to handle this rank
+		if (!allValid && firstHiNotValid <= x) {
+			// For each deprecated buckets
+			SortedMap<Integer, MutableRoaringBitmap> tailMap = hiToBitmap.tailMap(firstHiNotValid);
+
+			for (Map.Entry<Integer, MutableRoaringBitmap> e : tailMap.entrySet()) {
+				int currentHigh = e.getKey();
+				int index = Arrays.binarySearch(sortedHighs, 0, sortedHighs.length, currentHigh);
```
blasd: Good

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,284 @@
...
+					}
+					sortedCumulatedCardinality[index] = previousCardinality + bitmap.getCardinality();
+
+					if (currentHigh == Integer.MAX_VALUE) {
+						allValid = true;
+						firstHiNotValid = currentHigh;
+					} else {
+						firstHiNotValid = currentHigh + 1;
+					}
+					if (e.getKey() > x) {
```
blasd: Good

blasd:
This code is a port of another implementation where this specific choice has been properly validated. If you could contribute relevant benchmarks, we would be able to validate the choice of not maintaining an array of cumulated cardinalities.

richardstartin:
This work takes an approach not taken by pilosa, CRoaring, or roaring-rs, which I think requires significant justification. Note that features _not_ implemented in this pull requested include removal, in place AND, in place OR and in place XOR: **all of which will be significantly complicated by the need to maintain cumulative cardinalities.**

I think cumulative cardinalities will turn out to be a bad idea as the feature set on this class fills out, so unless they are removed, I can only oppose the merge of this PR. I will dismiss my own review and allow maintainers @lemire and @gssiyankai to judge the relative merits of this approach.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fixed. Thanks!

blasd:
Richard, the PR is in very early stage of work. While the code-base is used elsewhere it fits my needs, much more needs to be done before this RoaringTreeMap is integrated in this project (I'm not even happy with the class name). Cumulated cardinalities shall be removed in proper due time if demonstrated as useless. For now, I rather focus on implementing an MVP.

richardstartin:
Sure. I think contributions are great but I think it's a bad idea -> dropping out of this one.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fixed

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fixed

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fixed

blasd:
The JMH project mvn installation fails with: https://metamx.artifactoryonline.com/metamx/pub-libs-releases-local/org/roaringbitmap/real-roaring-dataset/0.6.30-SNAPSHOT/real-roaring-dataset-0.6.30-SNAPSHOT.pom. Return code is: 409

Any hint?

gssiyankai:
Maven is complaining because it cannot find the artifact real-roaring-dataset. There is a dedicated module in the repo to build it. Or you can use the script `run.sh` in the jmh folder to make things easier.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
Kineolyan: Sorry, I meant add +1 to the latest element of cumulative sum cardinalities, since it is the only one invalid. The test is fast and so is the increment. And it avoids the whole ensure... loop

blasd: I'm lost. If x is the last of sortedHigh, having written in the bitmap dismissed this last cardinality. I do not want to skip ensureCumulatives for high==x==lastInSortedHigh as it needs to be recomputed.

firstHighNotValid is the lowest high for which the cumulative is invalid. Not the last one for which it is valid.

blasd:
One of the agent failed because of JAVA_HOME. But the build is essentially green.

gssiyankai:
The `oraclejdk7` build is failing because travis jobs are now running on Trusty: https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming?utm_source=twitter&utm_medium=notification
It is reasonable to remove this jdk from `.travis.yml`

lemire:
@gssiyankai 
I have removed openjdk7.

lemire:
New tests are running.

lemire:
Ok. So unit tests are passing!

blasd:
Yep ^^

src/main/java/org/roaringbitmap/longlong/LongIteratorFlyweight.java
```diff
@@ -0,0 +1,159 @@
...
+ * 
+ * For better performance, consider the {@link RoaringTreeMap#forEach} method.
+ * 
+ * @author Borislav Ivanov
+ **/
+public class LongIteratorFlyweight implements PeekableLongIterator {
+
+  @Override
+  public boolean hasNext() {
+    // TODO Auto-generated method stub
```
lemire: I'd clean up such automatically generated comments.

src/main/java/org/roaringbitmap/longlong/LongIteratorFlyweight.java
```diff
@@ -0,0 +1,159 @@
...
+
+package org.roaringbitmap.longlong;
+
+/**
+ * Fast iterator minimizing the stress on the garbage collector. You can create one reusable
+ * instance of this class and then {@link #wrap(RoaringTreeMap)}
+ * 
+ * For better performance, consider the {@link RoaringTreeMap#forEach} method.
+ * 
+ * @author Borislav Ivanov
```
lemire: I'd update this comment.

src/main/java/org/roaringbitmap/longlong/LongIteratorFlyweight.java
```diff
@@ -0,0 +1,159 @@
...
+    // TODO Auto-generated method stub
+    return null;
+  }
+
+  public void wrap(RoaringTreeMap bitmap_c) {
+    // TODO Auto-generated method stub
+    
+  }
+
+//  private int hs;
```
lemire: Probably best not to include lots of commented out code.

src/main/java/org/roaringbitmap/longlong/PeekableLongIterator.java
```diff
@@ -0,0 +1,90 @@
...
+   *     </code></pre>
+   *  
+   *  The benefit of calling advanceIfNeeded is that each such call 
+   *  can be much faster than repeated calls to "next". The underlying
+   *  implementation can "skip" over some data.
+   *  
+   * 
+   * @param minval threshold
+   */
+  public void advanceIfNeeded(int minval);
```
lemire: Should probably be long minval.

src/main/java/org/roaringbitmap/longlong/RoaringIntPacking.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+// Hidden as it holds implementation details for RoaringTreeMap. We may decide to change the logic
+// here, hence it should
+// not be used elsewhere
+class RoaringIntPacking {
+
+  /**
+   * 
+   * @param id a long to decompose into two integers
+   * @return an int holding 32 bits of information of a long. Typically the highest order bits.
```
lemire: This documentation could be clearer. This is something like ceil(id / 2^32) when id <0 and floor(i^d/2^32). It should be clear that the range of values for which high(x) = 0 holds true is larger than the range of values for which high(x) = 1 holds true. So there could be design issues here.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,750 @@
...
+
+    latestAddedHigh = null;
+  }
+
+  /**
+   * Add the value to the container (set the value to "true"), whether it already appears or not.
+   *
+   * Java lacks native unsigned integers but the x argument is considered to be unsigned. Within
+   * bitmaps, numbers are ordered according to {@link Integer#compareUnsigned}. We order the numbers
+   * like 0, 1, ..., 2147483647, -2147483648, -2147483647,..., -1.
```
lemire: probably needs an update

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,750 @@
...
+  private int high(long id) {
+    return RoaringIntPacking.high(id);
+  }
+
+  /**
+   * Returns the number of distinct integers added to the bitmap (e.g., number of bits set).
+   *
+   * @return the cardinality
+   */
+  public long getLongCardinality() {
```
lemire: Evidently, we have a possible problem. What is the max cardinality?

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
```diff
@@ -0,0 +1,750 @@
...
+
+  private int binarySearch(int[] array, int from, int to, int key) {
+    if (signedLongs) {
+      return Arrays.binarySearch(array, from, to, key);
+    } else {
+      return unsignedBinarySearch(array, from, to, key, RoaringIntPacking.unsignedComparator());
+    }
+  }
+
+  // From Arrays.binarySearch (Comparator). Check with ROaring Util.unsignedBinarySearch
```
lemire: Typo

lemire:
@Kineolyan Thanks.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fix coming

lemire:
@richardstartin 

> What concerns me most is storing the cumulative cardinalities as long[]. 

Ok. Here is my view on this, and let me know if you agree. It can be a performance and complexity issue, I agree. But it is not exposed to the public API. If it is a problem, it ought to be possible to handle this level of details using various instances (e.g., using inheritance).

lemire:
@richardstartin @blasd 

> The code looks good but I feel strongly that the implementation should be unsigned.

Can we please discuss this by putting the arguments down on the table? There is a motivation (MAT) that I am sure many of us do not know well. 

I have a pretty hard time to see why it matters in the sense that signed 64-bit integers have 63-bit unsigned integers in them, which is a freaking lot. You don't even have a 63-bit memory space on current kernels.

Setting aside aesthetic concerns, what are the practical implications?

My view, and maybe it is naive, is that as long as it is defined and documented, we do not really care.

Note that the current implementation suffers from a problem, maybe, related to the fact that the high bits are defined as ``id>>32``, the image of which does not divide the space evenly.

blasd:
@lemire, any reason why RoaringBitmap implements ImmutableBitmapDataProvider but not BitmapDataProvider? It appears to be lacking to keep BitmapDataProvider as value type in the Map and follow the commited RoaringTreeMap.or

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Fixed

lemire:
@blasd 

> any reason why RoaringBitmap implements ImmutableBitmapDataProvider but not BitmapDataProvider? It appears to be lacking to keep BitmapDataProvider as value type in the Map and follow the commited RoaringTreeMap.or

Historical oversight that I am fixing now.

blasd:
@richardstartin @lemire 
About the long[] holding cumulative cardinalities, it is definitely a complexity issue as it has been large source of bugs until now. About the performance aspect, I'm about to commit some JMH benchmarks, but some inconsistencies appear for now:

Benchmark                                                 Mode  Cnt  Score    Error   Units
Cardinality64Benchmark.getCardinalityWithCache_Big       thrpt    3  3.037 ±  0.677  ops/us
Cardinality64Benchmark.getCardinalityWithCache_Small     thrpt    3  3.102 ±  0.539  ops/us
Cardinality64Benchmark.getCardinalityWithoutCache_Big    thrpt    3  0.001 ±  0.001  ops/us
Cardinality64Benchmark.getCardinalityWithoutCache_Small  thrpt    3  0.001 ±  0.001  ops/us
Cardinality64Benchmark.selectLastWithCache_Big           thrpt    3  2.291 ±  0.371  ops/us
Cardinality64Benchmark.selectLastWithCache_Small         thrpt    3  2.590 ±  0.289  ops/us
Cardinality64Benchmark.selectLastWithoutCache_Big        thrpt    3  0.001 ±  0.001  ops/us
Cardinality64Benchmark.selectLastWithoutCache_Small      thrpt    3  2.413 ±  0.633  ops/us

Big/Small refers to the size of the Map. Calling .select takes advantage of the cache. But I do not understand why .getCardinality without the cache is so slow with the small case.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: I tried adding to confirm we can have a cardinality above (2L * Integer.MAX_VALUE + 3L), the first cardinality rejected by RoaringBitmap. But something seems wrong with Roaring64NavigableMap.add(long,long).

I'm unsure how to document properly the fact the cardinality is bounded by Integer.MAX_VALUE * RoaringBitmap cardinality. The TreeMap may even be able to grow to above Integer.MAX_VALUE...

lemire:
@blasd  I think you are taking the right approach to assessing the question regarding the cumulative cardinalities.

I don't think I need to tell you that a good strategy is to go with the simplest thing that will work, and then optimize from there.

src/main/java/org/roaringbitmap/longlong/RoaringIntPacking.java
blasd: The design issue might be related to the failing test with .add(0, 3L * Integer.MAX_VALUE). I need to focus on this

lemire:
Please sync with master as I have pushed an update.

src/main/java/org/roaringbitmap/longlong/LongIteratorFlyweight.java
blasd: I went lazy on this. I remove the class as I've just introduced to follow some ports of existing JMH benchmarks.

richardstartin:
@blasd @lemire 

I will not be using this code as it stands so my views are moot.

blasd:
About signed and unsigned longs: I added a boolean in the constructor to enable choose the expected behaviour (e.g. should -1 before before or after 0).

blasd:
@lemire I have an issue with RoaringBitmap.select. I would expect .select to accept a negative int as an unsigned int. Then, I would expect the following to pass: I add all integers from 0 to -1, and select the -1: i expect it to return -1.

    @Test
    public void testRoaringBitmap() {
      RoaringBitmap map = new RoaringBitmap();

      long maxForRoaringBitmap = RoaringIntPacking.toUnsignedLong(-1);
      map.add(0L, maxForRoaringBitmap);

      Assert.assertEquals(-1, map.select(-1));
    }

Any clue?

blasd:
@lemire I feel like the test describing my issue is rather 

    @Test
    public void testRoaringBitmap_SelectAboveIntegerMaxValuePlusOne() {
      RoaringBitmap map = new RoaringBitmap();

      long maxForRoaringBitmap = RoaringIntPacking.toUnsignedLong(-1) + 1;
      map.add(0L, maxForRoaringBitmap);

      Assert.assertEquals(maxForRoaringBitmap, map.getLongCardinality());
      Assert.assertEquals(-1, map.select(-1));
    }

blasd:
@lemire I merged master. RoaringBitmap requires jdk7 so can not used on Integer.toUnsignedLong

I introduced RoaringIntPacking.toUnsignedLong in my PR. I guess it should go in the existing Util method

blasd:
@lemire What would be next steps before merging of the PR? There was an issue with the package. Roaring64NavigableMap can be configured to rely on either RoaringBitmap or MutableRoaringBitmap. In which package should I move this code?

What is the minimal set of boolean logic methods relevant to be implemented? (for now, I wrote only .or)

lemire:
@blasd 

Please sync your fork with master. We may have a little bit of a conflict (1 line?). I'm fine if you overwrite my code.

> What would be next steps before merging of the PR?

It looks like you have been doing lots of hard work. I'd push a little harder (see my comments below), though there is definitively some wiggle room as far as deciding what is necessary.

> There was an issue with the package. Roaring64NavigableMap can be configured to rely on either RoaringBitmap or MutableRoaringBitmap. In which package should I move this code?

Can you elaborate a little bit?

> What is the minimal set of boolean logic methods relevant to be implemented? (for now, I wrote only .or)
 

I think that just union is too little. I think we want ``or`` (union), ``and`` (intersection), ``xor`` (symmetric difference) and ``andnot`` (difference).  I suggest trying to implement all of them because...

1. It is easier than it sounds. Once you have implemented and/or, then the other ones are just variations. That is, the gap between just implementing unions and intersections, versus implementing the whole four operations is much smaller than you might intuitively think. It is not twice as long to implement four operations rather than just two.
2. I think it is a better test of the robustness of your approach to implement all four.

More comments:

- I would carefully document whether the serialization is robust... that is, if it is an unchanging documented format that we commit to... or whether the serialization is, as per the Java standard, so subject to changes as we modify the code. Both are probably fine, but Roaring uses a documented, fixed format that won't change and can be accessed by other programming languages. If you are using the standard Java serialization, then it needs to be made clear to the users.
- I'm guessing you have not implemented ``flip``. That's probably not too hard and worth doing.
- At the Roaring level, we urge users to call ``runOptimize()`` to improve the compression when they are done building the bitmaps. It seems that it would be a useful thing to support if you do not already.

lemire:
>  I merged master. RoaringBitmap requires jdk7 so can not used on Integer.toUnsignedLong

As you have now realized maybe, I caught this issue on my own and corrected it. Your fix is good too, maybe better.

lemire:
> I feel like the test describing my issue is rather

Feel free to add more unit tests, as needed, if you think I did not include enough.

blasd:
> There was an issue with the package. Roaring64NavigableMap can be configured to rely on either RoaringBitmap or MutableRoaringBitmap. In which package should I move this code?

A Roaring64NavigableMap over a RoaringBitmap can be instantiated through

    new Roaring64NavigableMap(true, new RoaringBitmapSupplier())

A Roaring64NavigableMap over a MutableRoaringBitmap can be instantiated through

    new Roaring64NavigableMap(true, new MutableRoaringBitmapSupplier())

The interface BitmapDataProviderSupplier enables provide any alternative BitmapDataProvider implementation.

blasd:
About .select for big integers, I will keep the additional unit-test. They were initially a way to specify my expected behavior. They remain an expression of the behavior expected by Roaring64NavigableMap over BitmapDataProvider.

blasd:
About serialization, current serialization process is clearly not portable. It depends on the underlying NavigableMap implementation which is not standard. We may think about a specification, like the one in RoaringBitmap. But I'm unsure what should be done about the tweak configuring if longs has to be considered as unsigned longs or not. I will document the fact the Serialization is not standard/portable for now, and it may (or even certainly given current state of work) change in the future. This will probably be improved in another PR, except if you feel it is mandatory int the initial implementation.

lemire:
I feel that most people would be satisfied if the issue re. serialization is clearly documented.

blasd:
> I think that just union is too little. I think we want or (union), and (intersection), xor (symmetric difference) and andnot (difference). I suggest trying to implement all of them because...

I'm not afraid at all ;). Implementing union were simply a way to demonstrate (to myself) current implementation let implementing such operation quite easily. But I do not want to implement what's optional before implementing the core functions. I could guess them, but your POV is much more relevant. And you may have observed some method added for RoaringBitmap should not have been introduced but are kept just for the sake of API stability.


About my MAT usecases, I need mainly memory compression, .select and .rank. This is why I did not focus on these other operations.

About flip, I'm about to add .flip(long). Should I also add .flip(long, long) ?

blasd:
@lemire would it be relevant to let .add returning true if the operation has changed the data? Like for Collection.add? In my case, it may help not invalidating the cardinality cache.

lemire:
Please see checkedAdd.

src/main/java/org/roaringbitmap/longlong/LongBitmapDataProvider.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Return the jth value stored in this bitmap.
+   *
+   * @param j index of the value
+   *
+   * @return the value
+   */
+  @Override
+  public long select(long j);
```
blasd: I just removed it, but it was initially coming from BitmapDataProvider:

@lemire Is there a reason for select to be overriden by BitmapDataProvider ?

lemire: @blasd  I think you  may have caught another oversight.

lemire:
@blasd  Please sync with master (minor release).

blasd:
@lemire What would is left before integration of the PR? I feel one thing is related to RoaringIntPacking.high and RoaringIntPacking.low

> This documentation could be clearer. This is something like ceil(id / 2^32) when id <0 and floor(i^d/2^32). It should be clear that the range of values for which high(x) = 0 holds true is larger than the range of values for which high(x) = 1 holds true. So there could be design issues here.

I am unsure to grasp the design issue you mentioned. Could you sketch a unit-test describing what could go wrong?

lemire:
@blasd 

I cleared out my comments. I don't recall exactly what I had in mind, but I was worried that without clear definitions, we would end up having what I call design issues.

Your code looks good. You are capturing the most significant 32 bits and the least significant 32 bits, and that's what the doc says. And it is clearly sound.

src/main/java/org/roaringbitmap/longlong/RoaringTreeMap.java
blasd: Hello. I feel this issue is now fixed. Could you confirm this? Thanks

blasd:
@Kineolyan Do you have more change requests? I feel I have fixed all your propositions.

src/main/java/org/roaringbitmap/longlong/MutableRoaringBitmapSupplier.java
```diff
@@ -0,0 +1,14 @@
+package org.roaringbitmap.longlong;
+
+import org.roaringbitmap.BitmapDataProvider;
+import org.roaringbitmap.buffer.MutableRoaringBitmap;
+
+public class MutableRoaringBitmapSupplier implements BitmapDataProviderSupplier {
+  private static final long serialVersionUID = -1332258289943674375L;
```
gssiyankai: I don't think this is necessary as the class is never serialized.

src/main/java/org/roaringbitmap/longlong/RoaringBitmapSupplier.java
```diff
@@ -0,0 +1,14 @@
+package org.roaringbitmap.longlong;
+
+import org.roaringbitmap.BitmapDataProvider;
+import org.roaringbitmap.RoaringBitmap;
+
+public class RoaringBitmapSupplier implements BitmapDataProviderSupplier {
+  private static final long serialVersionUID = -1332258289943674375L;
```
gssiyankai: Same as my previous comment.

src/main/java/org/roaringbitmap/longlong/MutableRoaringBitmapSupplier.java
blasd: Fixed. Thanks

src/main/java/org/roaringbitmap/longlong/RoaringBitmapSupplier.java
blasd: Fixed. Thanks

blasd:
I changed the default behavior to handle longs as unsigned longs by default.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+        int insertionPosition = -position - 1;
+        return insertionPosition;
+      }
+    } else {
+      // For each deprecated buckets
+      SortedMap<Integer, BitmapDataProvider> tailMap =
+          highToBitmap.tailMap(firstHighNotValid, true);
+
+      // TODO .size on tailMap make an iterator: arg
+      int indexOk = highToBitmap.size() - tailMap.size();
```
Kineolyan: Cannot you look for the index of firstHighNotValid instead, if you want to avoid the call the other iterator.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+        return mid; // key found
+      }
+    }
+    return -(low + 1); // key not found.
+  }
+
+  private void ensureOne(Map.Entry<Integer, BitmapDataProvider> e, int currentHigh, int indexOk) {
+
+    // sortedHighs are valid only up to some index
+    assert indexOk <= sortedHighs.length;
```
Kineolyan: Could you add a message to the assert, with at least `indexOf` and `sortedHighs.length` ? If it came to fail, this would be helpful more than just knowing the line.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+    } else if (indexOk < sortedHighs.length) {
+      index = -indexOk - 1;
+    } else {
+      index = -sortedHighs.length - 1;
+    }
+    assert index == binarySearch(sortedHighs, 0, indexOk, currentHigh);
+
+    if (index >= 0) {
+      // This would mean calling .ensureOne is useless: should never got here at the first time
+      throw new IllegalStateException("Unexpectely found " + currentHigh + " in "
```
Kineolyan: Typo: "Unexpecte-**d**-ly ..."

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+        return position + 1;
+      } else {
+        // This high has no bitmap: it could be between 2 highs with bitmaps
+        int insertionPosition = -position - 1;
+        return insertionPosition;
+      }
+    } else {
+      // For each deprecated buckets
+      SortedMap<Integer, BitmapDataProvider> tailMap =
+          highToBitmap.tailMap(firstHighNotValid, true);
```
Kineolyan: Can you consider combining `#tailMap(...)` to `#headMap(...)` so you only consider the relevant portion of the map ?
This requires that you have another way to compute `indexOf`. 
But the advantage is that you can iterate in reverse order and predict once for all how much you may need to grow `sortedHighs`

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+        } else {
+          ensureOne(e, currentHigh, indexOk);
+
+          // We have added one valid cardinality
+          indexOk++;
+        }
+
+      }
+
+      if (highToBitmap.isEmpty() || indexOk == highToBitmap.values().size()) {
```
Kineolyan: Do you really need to call `#size()` on the collection of values instead of calling it on `highToBitmap` ?

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+    invalidateAboveHigh(startHigh);
+  }
+
+  @Override
+  public LongIterator getReverseLongIterator() {
+    return toIterator(highToBitmap.descendingMap().entrySet().iterator(), true);
+  }
+
+  @Override
+  public void add(long x) {
```
Kineolyan: Can this replace the method `#addLong(...)`, as this one comes from the provider interface ?

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
```diff
@@ -0,0 +1,1351 @@
...
+      } else if (lowBitmap instanceof MutableRoaringBitmap) {
+        ((MutableRoaringBitmap) lowBitmap).flip(low);
+      } else {
+        // Fallback to a manual flip
+        if (lowBitmap.contains(low)) {
+          lowBitmap.remove(low);
+        } else {
+          lowBitmap.add(low);
+        }
+      }
```
Kineolyan: This branch of `#flip(...)` does not call `#invalidateAboveHigh(...)`, leading to invalid cardinalities and more.
See:

```java
  @Test
  public void testMe() {
    final Roaring64NavigableMap r = newUnsignedHeap();
    final long value = 1L;
    r.add(value);
    Assert.assertEquals(1, r.getLongCardinality());

    r.flip(1);
    Assert.assertEquals(0, r.getLongCardinality());
  }
```

Kineolyan: Also, I suggested earlier in the conversation that we could do a quick optimization when update the latest sortedHighs. 
I was thinking about something like this - idea applied to the manual part of flip.
```java
        // Fallback to a manual flip
        boolean contains = lowBitmap.contains(low);
        if (contains) {
          lowBitmap.remove(low);
        } else {
          lowBitmap.add(low);
        }

        if (allValid && sortedHighs.length > 0 && high == sortedHighs[sortedHighs.length - 1]) {
          // Directly update the cumulative count
          sortedCumulatedCardinality[high] += contains ? 1 : -1; 
        }
```
But I understand it requires that `add`, `remove` and `flip` return a boolean to indicate the results of the operations, which is not currently the case. I would understand if you want to keep that for another dev.

blasd: I considered returning true for add, remove and flip. @lemire provided an alternative to .add which returns a boolean but it seemed slightly more expensive. I need to check if it exists for .remove and .flip. I decided to keep the discussion around this for a another PR.

About maintaining manually the cardinalities, we maintain them even if high is not the last high (by increasing all greater high). Needing more thinking and opinions, I keep thinking this should be reserved for another PR. Thanks for pushing it any way.

Kineolyan: That seems a very good choice. I was mentioning it not to miss anything in the dev. But as the needed methods are not there yet, it is better to wait for the next iteration on that component.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: I'm uncomfortable with LongBitmapDataProvider.add. One may be confused with the actual meaning of .add(-1) which has a different meaning for BitmapDataProvider and for LongBitmapDataProvider.

What do you thing about rather naming LongBitmapDataProvider.addLong and LongBitmapDataProvider.removeLong ? Then, I would keep Roaring64NavigableMap.addInt (but out of the interface), and remove Roaring64NavigableMap.add

The followingtest fails on the last row, and demonstrate the confusive behavior:

    @Test
	public void testDefaultBehaviorLikeRoaring_MinusOneAsInt() {
		Roaring64NavigableMap longBitmap = newDefaultCtor();

		longBitmap.add(-1);
		bitmap.add(-1);

		// Ok as -1 === -1
		Assert.assertEquals(bitmap.select(0), longBitmap.select(0));

		// But RoaringBitmap.select has to be interpreted as an unsigned integer: not as an unsigned
		// long
		Assert.assertEquals(Util.toUnsignedLong(bitmap.select(0)), longBitmap.select(0));
	  }

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: Clearly useless

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: Fixed

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: Improved

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: I feel the relevant portion of actually the tail: we need to iterate along the tail to compute the missing cardinalities. Then I guess I miss your point here.

Anyway, `sortedHighs` resizing could take in account `highToBitmap` size (which is free for a `TreeMap`), and not depends on `firstHighNotValid` at all. Thanks for pointing this out.

I have difficulties to find a way to compute indexOf differently. I added a note about the feasibility to compute it from sortedHighs. `tailMap.size()` should not be called anyway.

blasd:
> I must say I am not comfortable with the naming Bitmap...Provider for the interface, as it represents more the actual Bitmap structure, than a provider.

I agree, LongBitmapDataProvider sticked to the convention introduced by BitmapDataProvider.

I'm unsure if you are rather challenging the naming of BitmapDataProviderSupplier introduced in this PR. Again, it sticks to BitmapDataProvider with, as you remarked, the Supplier suffix. I guess all these interfaces may be renamed from BitmapDataProvider to BitmapData at some point.

Kineolyan:
@blasd if it was made to stick with an existing name, then we will definitely follow that convention.
Changing it would be the matter of another PR, if any as it may imply breaking changes.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
Kineolyan: Good idea 😄 . You can try as well `#lastEntry`, which will give you the very last value to insert, at the cost of going through the highest branches.
But this was just a suggestion. I guess you would have spent time on this if it appeared to be a bottleneck in your benchmarks.

lemire:
@blasd I trust that you recommend an immediate merge of the PR, correct?

blasd:
@lemire The PR is in a good state of work regarding API and stability: it can be merged. There is a few pending improvements (essentially regarding tests, assertions), some of them coming within a few days. As you seem to release very regularly, I guess you are OK to if I open another PR right away.

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
Kineolyan: You say improved but I don't see any change ...

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
Kineolyan: Not seeing the fix 😢

src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
blasd: The fix was waiting on another machine and my remote access seemed broken. I felt it should not prevent merging the PR. Anyway, I have finally been able to push the related fix.

blasd:
@lemire @Kineolyan Before the merge, do you agree LongBitmapDataProvider.add and LongBitmapDataProvider.remove should be renamed to LongBitmapDataProvider.addLong and LongBitmapDataProvider.removeLong? I uncomfortable with the difference of behavior between LongBitmapDataProvider.add(-1L) and BitmapDataProvider.add(-1) while the interface suggests they are rigorously equivalent.

Kineolyan:
@blasd I don't see any reason to disagree. I would even say it is best to be sure not to be mistaken with `#add(int)` from `BitmapDataProvider`. Go for add/removeLong

lemire:
Ok! I'll make sure your name appears on the author list before release.

blasd:
I appreciate. Thanks!

PR #179, opened 2017-08-29
Introduce a parent pom

blasd:


blasd:
`mvn versions:set` fails in the root folder. However, this structure works in another project of mine for `mvn release:...` even if called in the root folder (at the difference the root module have packaging=pom in my case, but it should make no difference).

lemire:
Travis fails to with...

```
[ERROR] Failed to execute goal on project jmh: Could not resolve dependencies for project org.roaringbitmap:jmh:jar:0.6.51-SNAPSHOT: Failed to collect dependencies at org.roaringbitmap:RoaringBitmap:jar:0.6.51-SNAPSHOT: Failed to read artifact descriptor for org.roaringbitmap:RoaringBitmap:jar:0.6.51-SNAPSHOT: Could not transfer artifact org.roaringbitmap:RoaringBitmapParent:pom:0.6.51-SNAPSHOT from/to pub-libs (https://metamx.artifactoryonline.com/metamx/pub-libs-releases-local): Failed to transfer file: https://metamx.artifactoryonline.com/metamx/pub-libs-releases-local/org/roaringbitmap/RoaringBitmapParent/0.6.51-SNAPSHOT/RoaringBitmapParent-0.6.51-SNAPSHOT.pom. Return code is: 409 , ReasonPhrase:Conflict.
```

What happens on your machine if you type

```
mvn javadoc:test-javadoc && mvn animal-sniffer:check && mvn checkstyle:check  &&  mvn clean test jacoco:report install -Dgpg.skip=true && mvn -f real-roaring-dataset/pom.xml clean install &&  mvn -f ./jmh/pom.xml test -DBITMAP_TYPES=ROARING_ONLY
```

c.c. @gssiyankai

lemire:
Thanks btw, I am sure we will figure it out.

gssiyankai:
The build is failing on travis because the root pom that has been recently added has never been installed in the local maven repository. 
To fix the build, just replace line 22 in `.travis.yml` with: 
```shell
script: mvn install -N -f parent/pom.xml && mvn javadoc:test-javadoc && mvn animal-sniffer:check && mvn checkstyle:check  &&  mvn clean test jacoco:report install -Dgpg.skip=true && mvn -f real-roaring-dataset/pom.xml clean install &&  mvn -f ./jmh/pom.xml test -DBITMAP_TYPES=ROARING_ONLY
```

lemire:
@blasd  I think you need to update the travis file on your end.

blasd:
Regarding .travis.yml, I allowed myself adding an install step. It is introduced to prepare the travis env: here, I run a `mvn install` on the parent pom.xml: it will install all jars in the local mvn repository, skipping all tests and checks (checkstyle, javadoc, ...). It shall fail only if the compilation is KO: I feel OK to break very early if the compilation is KO.

    install: "mvn clean install --quiet -DskipTests -Dcheckstyle.skip -Dmaven.javadoc.skip=true"

I also slightly changed the script step: I feel clean and install are not useful as we already cleaned in the travis install step.

lemire:
Looks fine. Let us see what travis thinks.

blasd:
Some related steps would be to move the checkstyle plugin conf to parent/pom.xml: it would then apply to all modules. The jmh modules have kind of diverged in term of style ;).

lemire:
That would be fine if someone is willing to do the work of reformatting the JMH code.

blasd:
I'll do it then.

lemire:
@blasd  Travis is unhappy.

blasd:
The last commit applies the checkstyle conf on all modules. I applied the stylesheet in the project. Some additional styles have to be enforced. But we lacks a lot of Javadoc comments. Should we exclude this rule for the JMH module ?

lemire:
@blasd We could exclude JMH from the style tests, but open a corresponding issue that could be addressed at a later time.

blasd:
Open https://github.com/RoaringBitmap/RoaringBitmap/issues/180 for checkstyle over jmh

blasd:
About the choice of moving the root module to a subFolder and moving parent as root folder: I propose to keep current way (i.e. parent pom in a subFolder) if your release process is unchanged (e.g. I guess you simply do `mvn release:...` and `mvn:deploy:...`). I expect your release process will not be impacted. If the parent pom being in a subFolder requires to run mvn:release in a sub folder (e.g. ./parent), I would advise moving files so that the release process keep being done in the top folder.

lemire:
The current release is mvn clean, mvn prepare, mvn prepare... except if we have a version jump, but these are rare. Will the latest change break the release process?

lemire:
... mvn perform ...

PR #183, opened 2017-08-31
Switch to eh-cache sizeof agent, maintain one dataset in memory for j…

blasd:
JMH local tests runs from 11 min to 3 minutes. I maintain a single dataset in memory to prevent need of re-parsing it for each bitmap implementation.

I also added the memory package in travis script. In local, an issue appeared inside sizeof agent. I switch the eh-cache agent which is much easier to wire (e.g. no need for -javaagent argument -> tests can be runned from an IDE without additional conf)


blasd:
Single travis job seems to take around 17 minutes, to compare to 22 minuts with current master

lemire:
Wow! This looks like nice progress. Let us review this.

lemire:
These look like much more reasonable timings to me!

PR #118, opened 2016-08-07
Performancetweaks/august2016

lemire:
This is the PR corresponding to this issue...

https://github.com/RoaringBitmap/RoaringBitmap/issues/117

PR #189, opened 2017-10-26
fixes #118 Move first() and last() to ImmutableBitmapDataProvider

yizishou:


lemire:
Looks good. I'll add you to the list of contributors.

Issue #191, opened 2017-11-30
Pre-compute cardinalities for faster .rank/.select


blacelle:
My original usecase leading to implementing Roaring64NavigableMap relies massively on .rank and .select

Then, current Roaring64NavigableMap implementation holds a cache for cumulated cardinalities. I improved further my use-case by caching the cumulated cardinalities of underlying RoaringBitmap:

    new Roaring64NavigableMap(true, () -> new FastRankRoaringBitmap());

With current FastRankRoaringBitmap implementation

    public class FastRankRoaringBitmap extends RoaringBitmap {
	private int[] highToCumulatedCardinality = null;

	@Override
	public void add(long rangeStart, long rangeEnd) {
		highToCumulatedCardinality = null;

		super.add(rangeStart, rangeEnd);
	}

	@Override
	public void add(int x) {
		highToCumulatedCardinality = null;

		super.add(x);
	}

	@Override
	public void add(int... dat) {
		highToCumulatedCardinality = null;

		super.add(dat);
	}

	@Override
	public long rankLong(int x) {
		if (highToCumulatedCardinality == null) {
			highToCumulatedCardinality = new int[highLowContainer.size()];

			if (highToCumulatedCardinality.length == 0) {
				return 0;
			}
			highToCumulatedCardinality[0] = highLowContainer.getContainerAtIndex(0).getCardinality();

			for (int i = 1; i < highToCumulatedCardinality.length; i++) {
				highToCumulatedCardinality[i] =
						highToCumulatedCardinality[i - 1] + highLowContainer.getContainerAtIndex(i).getCardinality();
			}
		}

		short xhigh = Util.highbits(x);

		int index = Util.hybridUnsignedBinarySearch(this.highLowContainer.keys, 0, this.highLowContainer.size(), xhigh);

		boolean hasBitmapOnIdex;
		if (index < 0) {
			hasBitmapOnIdex = false;
			index = -1 - index;
		} else {
			hasBitmapOnIdex = true;
		}

		long size = 0;
		if (index > 0) {
			size += highToCumulatedCardinality[index - 1];
		}

		long rank = size;
		if (hasBitmapOnIdex) {
			rank = size + this.highLowContainer.getContainerAtIndex(index).rank(Util.lowbits(x));
		}

		// assert rank == super.rankLong(x);

		return rank;
	}

	@Override
	public int select(int j) {
		if (highToCumulatedCardinality == null) {
[...]

PR #192, opened 2017-12-07
Faster .rank .select

blasd:
Hello,

This PR follow-up https://github.com/RoaringBitmap/RoaringBitmap/issues/191

The general idea is to cache cardinalities, cumulated through buckets. This yield better performance when doing a lot of .rank and .select operations over a RoaringBitmap which is not written anymore.

Raw performance results (see TestRoaringBitmap_FastRank):

    AddOnlyThenSelectOnly FastRankRoaringBitmap=207ms
    AddOnlyThenSelectOnly RoaringBitmap=6064ms

    AddSelectRemoveAreMixed FastRankRoaringBitmap=1055ms
    AddSelectRemoveAreMixed RoaringBitmap=375ms

I am not fully-convinced `FastRankRoaringBitmap` is a proper naming for this class

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/FastRankRoaringBitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/FastRankRoaringBitmap.java
@@ -19,6 +19,10 @@ import java.util.Arrays;
  *
  */
 public class FastRankRoaringBitmap extends RoaringBitmap {
+  // Used to mark highToCumulatedCardinality as being dismissed: its cardinalities have to be
+  // recomputed
+  private static final int DISMISSED_MARKER = -1;
+
   // The cache of cardinalities: it maps the index of the underlying bucket to the cumulated
   // cardinalities (i.e. the sum of current bucket cardinalities plus all previous bucklet
   // cardinalities)
@@ -26,7 +30,15 @@ public class FastRankRoaringBitmap extends RoaringBitmap {
 
   private void resetCache() {
     // Reset the cache on any write operation
-    highToCumulatedCardinality = null;
+    if (highToCumulatedCardinality != null && highToCumulatedCardinality.length >= 1) {
+      // We tag the first bucket to indicate the cache is dismissed
+      highToCumulatedCardinality[0] = DISMISSED_MARKER;
+    }
+  }
+
+  // VisibleForTesting
+  boolean isCacheDismissed() {
+    return highToCumulatedCardinality == null || highToCumulatedCardinality[0] == DISMISSED_MARKER;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


blasd:
I changed the way to mark the cardianlity cache as dismissed. It enables re-using the int[] used as cache. The performance seems not improved but it will lower the GC pressure when doing a lot of consecutive .add/.select

    AddSelectRemoveAreMixed LargeInts FastRankRoaringBitmap=937ms
    AddSelectRemoveAreMixed LargeInts RoaringBitmap=387ms
    AddSelectRemoveAreMixed SmallInts FastRankRoaringBitmap=732ms
    AddSelectRemoveAreMixed SmallInts RoaringBitmap=625ms
    AddOnlyThenSelectOnly FastRankRoaringBitmap=184ms
    AddOnlyThenSelectOnly RoaringBitmap=5696ms

lemire:
@blasd 

I understand what you did and why you did it, and I am fine with merging it if you recommend merger. However, I have to ask...

I find it awkward to use a special value located in the first component of the vector to determine whether the cache is valid. It is not entirely intuitive, though it is clear enough.

Couldn't we, more simply, have a boolean flag. 

So instead of 

```
highToCumulatedCardinality == null || highToCumulatedCardinality[0] == DISMISSED_MARKER;
```

just have...

```
boolean isCacheDismissed;
```

The downside is that we add about a word of storage to store the boolean value, but that's probably negligible. It may even end up improving performance in that it is cheaper to check one flag than doing an array access.

I don't think it is a big deal, but it is a matter of elegance/simplicity.

blasd:
Hello,
I am completely fine to rely on a dedicated boolean.

Current implementation extends RoaringBitmap. Then, it does not work with `ImmutableRoaringBitmap`. It would require much less duplicated code to handle both case (and possibly part of `Roaring64NavigableMap`) if `RoaringArray` and `PointableRoaringArray` shared a common interface (e.g. `getContainerAtIndex(int i)` ; and the same for `Container.getCardinality and MappeableContainer.getCardinality()`). These additional interfaces would enable `FastRankRoaringBitmap `to be a decorator over underlying XXX (XXX being another interfaces as I guess would we not add `.getContainerAtIndex` to `ImmutableBitmapDataProvider`).

As these modifications requires many changes to existing interfaces (or changes to existing one), I guess it should be done in a later step.

lemire:
Ok. So let us be clear: you recommend as-is merge and release... but you’ll consider improvements later.

Is that correct?

blasd:
It can be merged as-is. I am not fully satisfied with the naming but if you
are ok with it, we are good to go.

I am also ready to try to implement the suggested improvments (soon enough
to differ merging) but I need your feeling about it. Anyway, even if you
are ok with the suggested refactoring, it may be preferable to do first an
intermediate (but still functional) release.

Le 13 déc. 2017 2:05 PM, "Daniel Lemire" <notifications@github.com> a
écrit :

Ok. So let us be clear: you recommend as-is merge and release... but you’ll
consider improvements later.

Is that correct?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/RoaringBitmap/RoaringBitmap/pull/192#issuecomment-351385282>,
or mute the thread

blasd:
The introduction of described interfaces seems somewhat related to https://github.com/RoaringBitmap/RoaringBitmap/issues/196, as they may help mixing different Containers implementation inside some `HybridRoaringBitmap `

lemire:
I'll merge and release soon.

lemire:
Merged. Sorry for the delay.

PR #195, opened 2017-12-09
fix unsigned equals bug in runcontainer

richardstartin:
This fixes a bug I encountered recently in equals between a RunContainer and an ArrayContainer whenever there is a run covering more than 15 bits.


lemire:
Looks good to me. Will be part of the next release.

lemire:
Released with 0.6.53.

PR #198, opened 2017-12-16
Correct release process

Kineolyan:
Hello all,

I can offer you the following changes to support back the old release process.
It added a new profile `release` that deactivates tests for _jmh_, _memory_ and _real-data-set_, although I am not sure it we should deactivate the last.
The profile is automatically activated by the configuration of maven-release-plugin.
And it also restores the old release tag name.

I did not tested the full release process, only to `release:prepare`, not to commit the real files.
I let you decide how you want to ensure it works completely.

By the way, you can reduce your command by using the following `mvn release:clean release:prepare release:perform`.

Final note: I could have added another profile for the perform phase not to run again the tests. But as you did not mention it, I assume it was not bothering you so I let it that way :smile:

### DIFF
--- a/jmh/pom.xml
+++ b/jmh/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.roaringbitmap</groupId>
         <artifactId>RoaringBitmapParent</artifactId>
-        <version>0.6.53</version>
+        <version>0.6.54-SNAPSHOT</version>
     </parent>
     <artifactId>jmh</artifactId>
     <packaging>jar</packaging>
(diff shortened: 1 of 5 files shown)


blasd:
It seems good to me. +1 to the proposed release command line

lemire:
Sadly, it does not change the result.

lemire:
As I document elsewhere, we needed the stagingProfileId value which explains why your PR did not fix my release woes... though I trust that your PR helped in other ways.

>  I could have added another profile for the perform phase not to run again the tests.

That would be very nice. How would you go about it?

PR #202, opened 2017-12-19
Only process modules to release when releasing

Kineolyan:


lemire:
Thanks. I understand what this should do, but I'll let @blasd review this.

blasd:
Thanks a lot.

    mvn install -DskipTests -> 5 modules
    mvn install -DskipTests -Prelease -> 2 modules

It is OK. However, we make -Prelease to be added also at release:perform: it will prevent the release from upgrading poms of all modules (which is part of the release process). To skip tests in release:perform, I introduced another profile: releasePerform. I feel like we just need to change the profile with a restricted set of modules from release to releasePerform (as the profile release is called for both release:prepare and release:perform).

blasd:
Even with me additional commit, I feel this will not work. More details tomorrow...

blasd:
The issue now is during release:prepare : we should not see all modules as default is not added (as there at least one other profile explicitly added). I guess the way out is to ensure no default is explicit during mvn:prepare (so that default profile applies and all modules are available) and add releasePerform profile during release:perform.

This means we should keep all explicit profiles out of the way of the default target ()

About 0.6.59: https://github.com/RoaringBitmap/RoaringBitmap/commit/c01955b759c0f367d124ef6403ea78ea41984b67 The release process updates the pom of only Parent and RoaringBitmap, not all modules

blasd:
The latest commit:

release:perform skip all tests and sees only Parent and RoaringBitmap
release:prepare plays all tests with restricted set for JMH (took about 6 minutes, I guess it is the goor order of magnitude)
mvn install -DskipTests sees all modules: OK

I guess you can try it @lemire

lemire:
Got it. Will try tomorrow or today.

lemire:
@blasd It works well in terms of command lines. Whether the artifacts sync with the maven repository is another matter.

PR #206, opened 2018-01-31
Fuzz testing

richardstartin:
I upgraded to Java 8 and added some basic fuzz testing for various invariants against random bitmaps. This PR is here mostly to see the differential test coverage achieved, and in case anyone else has any opionions about randomised testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ sudo: false
 jdk:
   - oraclejdk9
   - oraclejdk8
-  - openjdk7
 
 # we removed oraclejdk7 as per https://github.com/RoaringBitmap/RoaringBitmap/pull/176#issuecomment-322257170
 
--- a/pom.xml
+++ b/pom.xml
@@ -142,7 +142,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <javac.target>1.7</javac.target>
+        <javac.target>1.8</javac.target>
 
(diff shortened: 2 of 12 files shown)


richardstartin:
This PR may now be useful. Please consider it for review at convenience. 

A few points:

- If it's OK to drop JDK7, it will need to be removed from build config. 
- I don't think it's practical to do property/invariance based testing without functional language features. You need JDK8 to do this really. There may be drawbacks allowing functional features to leak into the main package however.
- There's a trade off between number of cases, sizes of bitmap, and build time. It might be better to be more exhaustive and put the fuzzer under a category to be activates by a profile. The fast profile would just run the unit tests
- This has already found a few bugs related to unsigned interpretation of signed shorts/ints in well covered code. It's a useful way to run tests.

lemire:
This looks good to me. More than good. And given that it is essentially just more and better testing, it seems that merging this is important.

Before taking the decision of dropping JDK 7 in future releases, I am opening the floor for discussions...

https://github.com/RoaringBitmap/RoaringBitmap/issues/207

richardstartin:
Just spotted a nasty bug in FastAggregation

lemire:
Ok. We are moving to JDK 8 so this is getting merged.

Great work.

richardstartin:
No problem. Actually I was in the middle of working on another unsigned error I found. I'll rebase and put in another PR.

lemire:
I think it is useful to have smallish PRs. Plus you get more (virtual) credit for your work!

PR #208, opened 2018-02-01
Changes to jmh/run.sh

maciej:
In this PR I introduce a couple of changes jmh/run.sh:
* running the `clean` Maven target is now optional. Explicitly pass `-clean` as the first argument to revert to old behavior. This cuts down on tool start time.
* skip check style. I don't think there is a point in having checkstyle run for this tool.
* add ability to pass an arbitrary number of arguments to JMH. This allow for easier passing of JMH parameters, like `-p dataset=dimension_003`

### DIFF
--- a/jmh/run.sh
+++ b/jmh/run.sh
@@ -1,24 +1,36 @@
 #!/bin/bash
 BASEDIR=$(dirname $0)
 
+CLEAN=""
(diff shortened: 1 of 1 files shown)


lemire:
This seems all good. Let us merge!

PR #209, opened 2018-02-01
fix unsigned error in runcontainer contains runcontainer

richardstartin:
This fixes a bug in RunContainer.contains(RunContainer) (this was my own handiwork unfortunately) which you would need a truncation of an identical run container to reproduce.


lemire:
Great.

PR #214, opened 2018-02-12
Bitmap add range performance

ppiotrow:
The code from benchmark is standard way in my codebase to create big bitmap representing the `Universe`, which is later it is used for negations.
The code creating `rangeOfOnes` was using the `iadd` code which is too complicated for this simple case. Instead I've switched to constructors including new one for `RunContainer`
The speed improvement is from `8k ops/s` to `12k ops/s` for standard bitmaps and `7.7k ops/s` to `10.5k ops/s` for `buffered`.

### DIFF
--- a/jmh/src/main/java/org/roaringbitmap/BasicBenchmark.java
+++ b/jmh/src/main/java/org/roaringbitmap/BasicBenchmark.java
@@ -159,8 +159,15 @@ public class BasicBenchmark {
   }
 
   @Benchmark
-  public RoaringBitmap createBitmapRange() {
+  public RoaringBitmap createBitmapRange_standard() {
     RoaringBitmap r = new RoaringBitmap();
+    r.add(0L, 300_000_000L);
+    return r;
+  }
+
+  @Benchmark
+  public MutableRoaringBitmap createBitmapRange_mutable() {
+    MutableRoaringBitmap r = new MutableRoaringBitmap();
     r.add(0L, 300_000_000L);
     return r;
   }
--- a/jmh/src/main/java/org/roaringbitmap/BasicBenchmark.java
+++ b/jmh/src/main/java/org/roaringbitmap/BasicBenchmark.java
@@ -159,6 +159,13 @@ public class BasicBenchmark {
   }
 
   @Benchmark
+  public RoaringBitmap createBitmapRange() {
+    RoaringBitmap r = new RoaringBitmap();
+    r.add(0L, 300_000_000L);
+    return r;
+  }
+
+  @Benchmark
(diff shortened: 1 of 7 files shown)


richardstartin:
@ppiotrow Just a thought - is this worth special casing? Yours is a common pattern e.g. [this](http://www.comp.nus.edu.sg/~chancy/sigmod98.pdf) relies on an "existence" bitmap. `RoaringBitmap.universe(long cardinality)` could create one container, and point all the keys to that container. It might be confusing if people start flipping bits in their universe bitmap, but that's what javadoc's for IMHO.

ppiotrow:
`RoaringBitmap.universe` might be good idea. You basically describe that `RB.add(final long rangeStart, final long rangeEnd)` is not optimal for `Universe` creation, two reasons for that:
1. Allocate lots of full RunContainers on heap.
2. Frequent `RoaringArray` resizing.
Regarding 2) I think we may come up with some idea.
Anyway, API design is harder than just merging this PR, which improves performance of existing public method, which is also used internally.
Personally I wouldn't like to end-up with bitmap that would behave differently as the contract states (shared containers).

ppiotrow:
I've replaced `Run/ArrayContainer.serializedSizeInBytes` with precalculated threshold, but got failing tests. It appears`RoaringArray.serializedSizeInBytes` uses containers methods `getArraySizeInBytes` which gives different results for some containers. Is this expected?

lemire:
@ppiotrow Can you elaborate?

ppiotrow:
@lemire https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java#L565
I would expect that the method `RoaringArray.serializedSizeInBytes()` calls method `serializedSizeInBytes()` on underlying containers. For some unclear to me reason it calls `getArraySizeInBytes`.
For `ArrayContainer` getArraySizeInBytes() is different than `serializedSizeInBytes()`, why?

lemire:
I don’t know if it is still the case, but it used to be that you could just serialize one container alone. For Array, you need the cardinality but only when it is serialized alone... when serialized as part of the roaring bitmap, we omit the cardinality.

If I got my story right... there should be two bytes of difference.

ppiotrow:
Ok, it wasn't clear to me and I accidentally changed the behaviour of previous code by invalid assumption. Right now everything is just same as it used to be, but faster. Ready to be merged if no review issues.

lemire:
Seems like a no-brainer to merge. Sure.

This will be part of our next release.

PR #215, opened 2018-02-13
Trim() is not releasing all unused bytes

ppiotrow:
The RoaringArray's `keys` and `values` arrays are extended with some extra space according to [code](https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java#L320).
The space is not recovered after `RoaringBitmap.trim()` call. This could cause huge heap usage while building millions of bitmaps.
The unit tests shows how to easily waste `5292 bytes` per one bitmap.


lemire:
+1

lemire:
Will be part of next release.

PR #218, opened 2018-02-14
.toString method inconsistent between containers

ppiotrow:
For `RunContainers` the test fails with
```
org.junit.ComparisonFailure: 
Expected :{0,1,17,32767,-3,-1}
Actual   :{0,17,32767,65533,65535}
```
as the `toString()` for `RC` returns `[0,1][17,17][32767,32767][65533,65533][65535,65535]`
This was very misleading for me during my first pull requests.
The purpose of this PR is to decide what is the correct convention of `toString()` method or argumentation why there should be an exception.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
@@ -1104,10 +1104,10 @@ public final class ArrayContainer extends Container implements Cloneable {
     StringBuilder sb = new StringBuilder();
     sb.append("{");
     for (int i = 0; i < this.cardinality - 1; i++) {
-      sb.append(this.content[i]);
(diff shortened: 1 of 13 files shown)


ppiotrow:
I vote for containers `String` representation to be between 0 and 65 536. This is consistent with [paper](https://arxiv.org/pdf/1603.06549.pdf).
`Roaring partitions the space [0,n) into chunks of 216 integers([0,2^16),[2^16,2×2^16),...)`.
This would be also consistent with RoaringBitmap `toString()` which yields values from 0 to 4294967296.
This means the `toString` implementation of Array and Run containers need to be changed. WDYT?

ppiotrow:
Ready to merged. Also there was bug in here https://github.com/RoaringBitmap/RoaringBitmap/pull/218/files#diff-29bec97d8e6f823dc718eb71d6e6e806R2277

PR #219, opened 2018-02-14
make getLength and getValue public

richardstartin:
I have found that if calling code "knows" about runs, it's much easier to make the code as fast as possible. I imagine there will be some pushback on this. This is a small, low risk change, but might evoke strong feelings from a software engineering point of view.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
@@ -1001,7 +1001,14 @@ public final class RunContainer extends Container implements Cloneable {
     return sum;
   }
 
-  short getLength(int index) {
+  /**
+   * Gets the length of the run at the index.
+   * @param index the index of the run.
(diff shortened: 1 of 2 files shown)


lemire:
Thanks!

richardstartin:
Thoughts?

lemire:
I have no problem with this code, but would you be so kind as to add documentation to these methods. Because these functions become public, it should be expected that users can find documentation.

If you can't, somehow, document them from your end, I'll merge and document.

richardstartin:
I will add some JavaDoc this weekend

richardstartin:
Done (fixed up onto the original commit).

ppiotrow:
For consistency please add public to `MappeableRunContainer.java`.

richardstartin:
To be honest I'm only changing this in `RunContainer` because I actually use the class.

lemire:
I think it is now a no-brainer.

richardstartin:
Thanks for keeping me honest with the thorough review.

PR #224, opened 2018-03-07
port various optimisations to the buffer package

richardstartin:
This isn't necessarily for merge, just so @ppiotrow  and @maciej know it's here. This contains some of the improvements made to `ParallelAggregation` (copy and paste).


lemire:
@richardstartin Might be worthwhile to merge it, unless you think that it is unwise?

richardstartin:
It benchmarks better, but it's not actually going to solve either of the problems, having seen the statistics (but not the logical expressions which I glean contain intersections and unions). It can't hurt to merge.

PR #229, opened 2018-03-11
measure test coverage from perf tests - don't merge

richardstartin:
I just want to see the differential coverage.


richardstartin:
So the change if this sort of thing were moved to JMH is negligible.

Issue #225, opened 2018-03-07
Avoid double copy in in-place array OR


lemire:
This code in in-place OR within the ArrayContainer class involves a duplicate copy:

```
    if(totalCardinality >= content.length) {
      increaseCapacity(totalCardinality);
    }
    System.arraycopy(content, 0, content, value2.cardinality, cardinality);
```

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java#L658-L661

It also occurs in the iadd method. It is also possible that the remove method has duplicate copies.
We should hunt down such code.

amitdesai03:
Hi Lemire, I am new to open source and wanted to participate. My recent project involved heavy use of roaring bitmap and have been exploring internals of it for while. If its ok with you, i would love to take stab at fixing this.

lemire:
That's an excellent idea. Go ahead and prepare a pull request.

amitdesai03:
Appreciate your welcome gesture for newbies. I have submitted PR. Please review. I did run all tests and benchmarks, but not sure where to see performance to compare.

lemire:
@amitdesai03  Looking at your code, I don't think you are such a newbie.

amitdesai03:
@lemire thanks, I have been developer for decade now, just new to open source :)
Also really admire the work you guys have done with roaring. Lot to learn, just looking at it.

lemire:
@amitdesai03 

> I have been developer for decade now

Yes. That's more like what I expected.

> just new to open source

There is still much to improve in Roaring.

lemire:
I'll open a separate issue for iadd.

PR #230, opened 2018-03-16
Fix to avoid double copy in in-place array OR

amitdesai03:
**In case when we increased capacity, we used to follow below steps:**

1. increaseCapacity(Arrays.copyOf)   - Copy 1st time
2. System.arrayCopy to offset  - Copy 2nd time
3. Perform Union

**With changes, it saves 2 copies:**

1. - create new array. (No copy)
2. - Perform union with new array as destination

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,4 +20,5 @@ Olivier Peyrusse,
 Richard Startin,
 Erik Gorset,
 Benoit Lacelle,
-Yizhu Sun
+Yizhu Sun,
+Amit Desai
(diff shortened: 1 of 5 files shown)


lemire:
Looks good. Would you extend your work to the buffer class?

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableArrayContainer.java#L799-L802

It is mostly just copy-and-paste.

lemire:
Reference: https://github.com/RoaringBitmap/RoaringBitmap/issues/225

amitdesai03:
Sure, will do.

lemire:
Please also add your name to 

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/AUTHORS

lemire:
Or I can add your name later. But your name will need to be added.

amitdesai03:
@lemire added changes, please check

lemire:
Looks good. Will be part of the next release.

Issue #231, opened 2018-03-17
Avoid double copy in in-place array add


lemire:
This code in in-place add within the ArrayContainer class involves a duplicate copy:

    if (newcardinality >= this.content.length) {
      increaseCapacity(newcardinality);
    }
    System.arraycopy(content, indexend, content, indexstart + rangelength, cardinality - indexend);

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java#L461-L464

The same problem occurs in the corresponding buffer class.

This is a follow up to https://github.com/RoaringBitmap/RoaringBitmap/issues/225

amitdesai03:
This one was tricky to implement.
Here is gist which passes all tests.
I am not sure how to run benchmark against it.

https://gist.github.com/amitdesai03/495c99432c0b0c60987daaaa728f2d33

If you approve of it, i will remove all comments and submit clean PR.

richardstartin:
@amitdesai03 it might be better just to submit a PR anyway, the difference is clearer to see that way.

lemire:
Can you follow @richardstatin’s advice?

amitdesai03:
@richardstartin @lemire  absolutely, here you go:
https://github.com/RoaringBitmap/RoaringBitmap/pull/232

lemire:
Fixed by PR https://github.com/RoaringBitmap/RoaringBitmap/pull/232

PR #232, opened 2018-03-17
Proposed fix - Avoid double copy in in-place array add

amitdesai03:


amitdesai03:
Reference: https://github.com/RoaringBitmap/RoaringBitmap/issues/231

amitdesai03:
Ran iadd benchmarks against huge list of ranges:

**Before**

Benchmark         Mode  Cnt      Score      Error  Units
AddBenchmark.add  avgt    5  57538.034 ± 2756.394  us/op

**After**:

Benchmark         Mode  Cnt      Score      Error  Units
AddBenchmark.add  avgt    5  55455.495 ± 2795.335  us/op

lemire:
@amitdesai03 Please run ``mvn checkstyle:check`` on your code. It appears that you have style violation. We enforce a strict style. (It avoids having to deal with folks reformatting the code according to their favourite style.)

amitdesai03:
@lemire yes I understand, I actually checked in with lot of comments so as you can review. 
If i format it, it just makes comments reading harder. I was thinking that after your review, i will checkin with checkstyle complaint version. But if you want, i can checkin right away?

**Here is more accurate benchmark, which shows definite improvement by :**

Benchmark                  Mode  Cnt     Score     Error  Units
AddBenchmark.add           avgt   20  5712.009 ± 104.209  us/op
AddBenchmark.addOptimized  avgt   20  5684.183 ±  49.245  us/op

I am not sure if I should checkin benchmarking code, as it introduces new random range set to real data set directory?

lemire:
> I am not sure if I should checkin benchmarking code, as it introduces new random range set to real data set directory?

Can you elaborate? I am not sure I understand why you are hesitating to include your benchmarking code?

amitdesai03:
Since name represented real data set, what i have is generated data set based on randomization. Also have custom zip reader for ranges in contrast to what exists already.

I will check in all, you be the judge :), lemme know if needs some modification

lemire:
Good.

amitdesai03:
done, pls verify

lemire:
I think this solves it. Great work.

amitdesai03:
tx, learned quite few things like jmh and shortbuffer

PR #236, opened 2018-03-22
implement intersects with range

richardstartin:
Just knocked this up quickly. Let me know if there are obvious algorithmic improvements.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
@@ -670,6 +670,13 @@ public final class ArrayContainer extends Container implements Cloneable {
     return x.intersects(this);
   }
 
+  @Override
+  public boolean intersects(short minimum, short supremum) {
+    int pos = Util.unsignedBinarySearch(content, 0, cardinality, minimum);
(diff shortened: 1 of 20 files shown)


lemire:
Presumably, your code checks for intersections with the interval [min,max). Right? It is not clearly specified. But if so, then you have the problem that 1<<32 cannot be expressed. That's why we use longs elsewhere in the API.

(Recall that we interpret Java's ints as unsigned integers.)

lemire:
(Other than that, the code is superb.)

lemire:
You are correct and I was wrong, the specification is there... note however the typo:

>   param supremum the exlusive unsigned upper bound of the range

But my objection still stand since you can't capture the whole range [0,1<<32).

richardstartin:
I agree with you. I've just made those changes.

lemire:
In the past, people have felt strongly that the ranges should be checked. So we have a function called ``rangeSanityCheck``. Do you want to have a look?

lemire:
(I can merge and add the check later if you'd prefer. But I suspect that some people really want the check.)

richardstartin:
I'm not sure there's much left to do on this, I'm happy for it to be merged.

lemire:
I agree. This looks (very) excellent to me.

Issue #239, opened 2018-03-30
Replace cardinality checks with ``isEmpty`` calls


lemire:
To check that a container is not empty, we routinely do a cardinality check ``c.getCardinality() > 0``. We should replace those with ``! c.isEmpty()`` for clarity and performance.

amitdesai03:
Will club it as part of https://github.com/RoaringBitmap/RoaringBitmap/pull/240

Issue #185, opened 2017-09-12
Avoid unnecessary container allocations


lemire:
In many instances, especially when computing intersections, we generate empty containers that are immediately garbage collected.

There are clever tricks we could use to start scanning the input containers, advance up to the point where there might be values to be intersected, and bail out without generating an empty container if the intersection is empty.

amitdesai03:
@lemire  could you pls point me to code which you are referring to?

lemire:
I will document it better soon.

lemire:
Sorry for the delay. Here are code samples that could be improved:

```
        final Container c1 = x1.highLowContainer.getContainerAtIndex(pos1);
        final Container c2 = x2.highLowContainer.getContainerAtIndex(pos2);
        final Container c = c1.and(c2);
        if (c.getCardinality() > 0) {
          answer.highLowContainer.append(s1, c);
        }
```

```
        final Container c1 = x1.highLowContainer.getContainerAtIndex(pos1);
        final Container c2 = x2.highLowContainer.getContainerAtIndex(pos2);
        final Container c = c1.andNot(c2);
        if (c.getCardinality() > 0) {
          answer.highLowContainer.append(s1, c);
        }
```

```
        Container c =
            bm.highLowContainer.getContainerAtIndex(i).not(containerStart, containerLast + 1);
        if (c.getCardinality() > 0) {
          answer.highLowContainer.insertNewKeyValueAt(-j - 1, (short) hb, c);
        }
```

```
        final Container c = rb.highLowContainer.getContainerAtIndex(i).remove(lbStart, lbLast + 1);
        if (c.getCardinality() > 0) {
          answer.highLowContainer.append((short) hbStart, c);
        }
```

and so forth...

richardstartin:
Also, in those snippets, the cardinality will be evaluated for `RunContainers`. There is a method now `Container.isEmpty` which doesn't need to evaluate the cardinality.

lemire:
@richardstartin  Indeed. So there are multiple opportunities for optimization.

lemire:
@richardstartin For clarity I turned your point into a distinct issue: https://github.com/RoaringBitmap/RoaringBitmap/issues/239

amitdesai03:
Thanks @richardstartin @lemire 

Have created PR which does not create temp containers in case both compared containers are empty.

Is there more needed to this optimization?
If not, i will apply this change to all necessary places including replacing getCardinality with isEmpty

[...]

PR #240, opened 2018-03-30
Prevent creating temp containers

amitdesai03:
https://github.com/RoaringBitmap/RoaringBitmap/issues/185
https://github.com/RoaringBitmap/RoaringBitmap/issues/239

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -287,9 +287,11 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
       if (s1 == s2) {
         final Container c1 = x1.highLowContainer.getContainerAtIndex(pos1);
         final Container c2 = x2.highLowContainer.getContainerAtIndex(pos2);
-        final Container c = c1.and(c2);
-        if (c.getCardinality() > 0) {
-          answer.highLowContainer.append(s1, c);
+        if (!c1.isEmpty() && !c2.isEmpty()) {
+          final Container c = c1.and(c2);
+          if (!c.isEmpty()) {
+            answer.highLowContainer.append(s1, c);
+          }
         }
         ++pos1;
         ++pos2;


lemire:
I think that we do not store empty containers, so this condition will always be true. No?

lemire:
> I though we never confiscate empty containers on remove() call. So there is possibility.

That would be a bug. I don't think we do so... 

```
  public void remove(final int x) {
    final short hb = Util.highbits(x);
    final int i = highLowContainer.getIndex(hb);
    if (i < 0) {
      return;
    }
    highLowContainer.setContainerAtIndex(i,
        highLowContainer.getContainerAtIndex(i).remove(Util.lowbits(x)));
    if (highLowContainer.getContainerAtIndex(i).isEmpty()) {
      highLowContainer.removeAtIndex(i);
    }
  }
```

lemire:
> Also, would you be open to have gitter/irc channel for more agile communication?

We have a [mailing list as you know](https://groups.google.com/forum/#!forum/roaring-bitmaps).

We can create more channels. If someone organizes a chat channel, we will be sure to add links to it.

lemire:
> Did you mean we modify internals of and() methods to prevent allocating anything till something found to intersect?

That's what I meant, yes. See my description:

_There are clever tricks we could use to start scanning the input containers, advance up to the point where there might be values to be intersected, and bail out without generating an empty container if the intersection is empty._

I am thinking that instead of eagerly allocating a new container, we could delay the allocation until we we know the result won't be empty. I'd start with array-array intersections.

richardstartin:
I’d be tempted to verify the container isn’t scalarised before optimising this. If you get a significantly worse result and different allocation rates from -prof gc when disabling escape analysis when executing this section of code, it’s being optimised for you.

lemire:
> I’d be tempted to verify the container isn’t scalarised before optimising this. If you get a significantly worse result and different allocation rates from -prof gc when disabling escape analysis when executing this section of code, it’s being optimised for you.

That's a very good point. I'll recopy your comment as part of the main issue.

lemire:
I will close this for now.

PR #242, opened 2018-04-11
Evaluate batch iteration

richardstartin:
I have implemented a similar pattern to the go library for batch iteration. I will probably port this to the buffer package later this week. I would be interested in seeing benchmark results for this implementation.


richardstartin:
OK. I plan to port to the buffer version and have used the wrong email address on some commits which needs fixing (I will need to force push).

lemire:
@richardstartin I think we want to do something meaningful like sum the values being iterated over. And compare the old-school iterators with the new iterators.

Is this what you have in mind?

richardstartin:
@lemire The code I propose evaluating is on #243 - can we discuss there?

lemire:
Of course.

lemire:
I'm going to lock this conversation.

PR #253, opened 2018-06-04
implement nextSetBit

richardstartin:
This implements similar functionality to `java.util.BitSet.nextSetBit`, except the next bit is returned as a long, since RoaringBitmaps can contain all 32 bits.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
@@ -1142,6 +1142,13 @@ public final class ArrayContainer extends Container implements Cloneable {
   }
 
   @Override
+  public int nextSetBit(short fromValue) {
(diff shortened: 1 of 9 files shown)


lemire:
I'll merge and try to throw in some benchmarks.

lemire:
(Great work btw.)

richardstartin:
So it's multiplicatively slower because of container lookup overhead and binary searches. I didn't try to optimise this code at all though. Rather than implementing the bitset API verbatim it might have been interesting to get the next page of values after a given value, like a skip list.

richardstartin:
@lemire agreed that it's a terrible iteration strategy to purposefully forget where you are each time you go back to the data structure, but it potentially has interesting use cases, either to retrieve an iterator starting at an offset, or to fill an array of values after an offset. This would be useful for stateless pagination.

lemire:
@richardstartin 

If you retrieve more than one value each time, thus amortizing the search cost, then it makes a lot of sense, I agree.

Issue #254, opened 2018-06-12
RunContainer::contains(ArrayContainer) throws Exception due to off by one termination check


sesuncedu:
The while loop in this method has a termination condition checks for run indexes <= numberOfRuns(), which is incorrect. 

Additionally, the return value is computed incorrectly.

I have a patch for this- I just need to add a unit test (I probably ought to check the methods for the other container types as well).

Addendum:  The existing test does not trigger this error as the relevant check tests for containment with a putative subset larger than the putative set. This short circuits the precise containment logic.

Further addendum: Because the valuesLength array is sized to hold four runs, the code will not throw an exception unless the array is full (e.g. if bitmap is trimmed or if the valuesLength array is filled).

lemire:
Reopen if this is still an issue.

PR #255, opened 2018-06-12
Tests and fix for issue #254 (RunContainer.contains(ArrayContainer))

sesuncedu:
Legal-Meta - I didn't notice I didn't have sign-off set when I committed the change  to the test case, but by signing off on the substantive fix, and by submitting this pull request, it is my express intent to grant any and all rights necessary to incorporate these changes into the existing codebase, and that I am legally entitled to grant such rights under the laws of the United States of America and the State of North Carolina.
 
/s/ Simon Spero

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
@@ -809,11 +809,12 @@ public final class RunContainer extends Container implements Cloneable {
     }
     int ia = 0, ir = 0;
     while(ia < arrayContainer.getCardinality() && ir < runCount) {
-      int start = getValue(ir);
+      int start = toIntUnsigned(this.getValue(ir));
       int stop = start + toIntUnsigned(getLength(ir));
-      if(arrayContainer.content[ia] < start) {
+      int ac = toIntUnsigned(arrayContainer.content[ia]);
+      if(ac < start) {
         return false;
-      } else if (arrayContainer.content[ia] > stop) {
+      } else if (ac > stop) {
         ++ir;
       } else {
         ++ia;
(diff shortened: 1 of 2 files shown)


roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -808,7 +808,7 @@ protected boolean contains(ArrayContainer arrayContainer) {
       return false;
     }
     int ia = 0, ir = 0;
-    while(ia < arrayContainer.getCardinality() && ir <= runCount) {
+    while(ia < arrayContainer.getCardinality() && ir < runCount) {
       int start = getValue(ir);
```
richardstartin: There is actually yet another error here: this should be 

```java 
int start = Util.toIntUnsigned(getValue(ir));
```

richardstartin:
Good catch. There is another error in the method you are patching, perhaps you could add that fix in too while you are at it?

roaringbitmap/src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -808,7 +808,7 @@ protected boolean contains(ArrayContainer arrayContainer) {
...
       int stop = start + toIntUnsigned(getLength(ir));
       if(arrayContainer.content[ia] < start) {
```
richardstartin: There's another one here; we should be storing `arrayContainer.content[ia]` in an int variable:

```java
int value = Util.toIntUnsigned(arrayContainer.content[ia]);
if (value < start) {
...
} else if (value > stop) {
...```

richardstartin:
LGTM

lemire:
Taking into account  @richardstartin 's review... merging...

lemire:
Updated https://github.com/RoaringBitmap/RoaringBitmap/blob/master/AUTHORS

PR #256, opened 2018-06-12
simplify and fix signedness issues in contains methods

richardstartin:
There are a few opportunities to reduce duplication between the `Container::contains(Container)` methods and `Container.contains(int min, int sup)` methods, and this fixes some errors where signed values should be treated as unsigned. 

These changes have been made specifically to avoid any merge conflicts with #255.


sesuncedu:
I should have checked this before I  added the masking to the PR below :-)

lemire:
Looks great.

Issue #260, opened 2018-06-21
RoaringBitmap.addOffset causing problems with serialization/deserialization


jeremycod:
Hi,

Based on Issue #251 I've created a utility method that aggregates two bitmaps
 
```
  override def concatenateBitmaps(index: Int, resultBitSets: Array[BitSet], numRows: Int, offset: Int): Unit = {
    val shifted = RoaringBitmap.addOffset(underlying.toRoaringBitmap, offset)
    val underlyingBitSet = resultBitSets(index).underlying.asInstanceOf[MutableRoaringBitmap].toRoaringBitmap
    underlyingBitSet.or(shifted)
    resultBitSets(index) = new RoaringBitSet(underlyingBitSet.toMutableRoaringBitmap, numRows)
  }
```
 
Results of this operation are serialized and deserialized later again. Sometimes, it hits EOFException and I can't deserialize bitmap from the file. I was able to reproduce the problem in a simple unit test that I'm sending in attachment. 
It seems that the problem is in this part:

`val shifted = RoaringBitmap.addOffset(underlying.toRoaringBitmap, offset)`

With test data I've provided and offset 5950 shifted value that I get as a result has BitMapContainer with actual data and additional ArrayContainer which is empty. Even debugger shows exception when calling RoaringBitmap.toString(), but main issue is when I serialize and try to deserialize such bitmap, it always produces exception:

```
[info] RoaringBitSetsSerializationProblemSpec:
[info] - should serialize and deserialize the same roaring bitmap *** FAILED ***
[info]   java.io.EOFException:
[info]   at java.io.DataInputStream.readFully(DataInputStream.java:197)
[info]   at java.io.DataInputStream.readLong(DataInputStream.java:416)
[info]   at java.io.ObjectInputStream$BlockDataInputStream.readLong(ObjectInputStream.java:3163)
[info]   at java.io.ObjectInputStream.readLong(ObjectInputStream.java:1034)
[info]   at org.roaringbitmap.buffer.MutableRoaringArray.deserialize(MutableRoaringArray.java:295)
[info]   at org.roaringbitmap.buffer.MutableRoaringArray.readExternal(MutableRoaringArray.java:499)
[info]   at org.roaringbitmap.buffer.MutableRoaringBitmap.readExternal(MutableRoaringBitmap.java:1269)
[info]   at visier.platform.data.bitset.RoaringBitSet.deserialize(BitSet.scala:244)
[...]

PR #261, opened 2018-06-22
Tests and fixes for RoaringBitmap.addOffset

richardstartin:
This PR shows up some issues in `RoaringBitmap.addOffset` and fixes some low hanging fruit.

- There was an array index out of bounds in `RunContainer.append` when a run container was placed after a bitmap container
- There was an assertion error when a run container came after a bitmap container and `Util.cardinalityInRange` was called.
- An assertion has been added to `RoaringArray.append` to reject empty containers.
- A test for equivalence of shifted bitmaps before and after (de)serialization has been added. 
- Tests show some data is lost when the offset is not a multiple of 65536
- Remove some very slow fuzz tests for parallel aggregation because I can't be bothered to watch them run
- A test case for the data scenario reported in #260 (could not reproduce)

This PR will break because it contains tests which capture bugs but I don't have time  to fix them.


richardstartin:
The tests should all be passing now.

richardstartin:
I've ported these fixes to the buffer implementation just now (I didn't see any evidence of empty containers there either).

lemire:
LGTM. Merging.

Issue #72, opened 2016-01-28
Roaring bitmaps for matrix operations


cogmission:
Hi, I apologize if this isn't the right venue for asking this question?

But I was wondering if it wouldn't be ridiculous to implement a SparseMatrix using a collection of RoaringBitmaps as its rows?

What do you think?

Cheers,
David

lemire:
A bitmap/bitset can be thought of as a set of integers. The rows of a matrix are more like arrays. So there is no obvious match in my mind. Care to elaorate?

cogmission:
Hi Daniel, thank you for responding so quickly, I'm a big fan of your work!

The idea would be to have a Matrix with bits instead of integers (in my case). Each row would be a Roaring Bitmap which would save on memory? For my project, I'm using int[]s as the payload that gets passed around, and the integers are only either 1 or 0; the arrays in the usual case are 2048 bits wide with 2% sparsity (typically == 40 on bits).

I'm thinking about a way to minimize memory consumption without losing performance. Using 32 bits to represent a zero or one, you can see, is kind of a waste - however I did read somewhere that the JVM is more efficient using integers (having something to do with register sizes or something?) [Here](http://stackoverflow.com/questions/8804435/alternative-to-java-bitset-with-array-like-performance) is one place where I saw some discussion of this, though I can't find the original source.

lemire:
Oh! Yes, if it is a matrix of bits, then yes, it could work.

If you are using ``int[]`` to store 0s and 1s, you could reduce your memory usage by nearly a factor of 4 by using ``boolean[]`` arrays instead as a ``boolean`` in an array use as little as 1 byte in Java. 

It is entirely possible that you might do even better with Roaring bitmaps. In the abstract, it is a bit difficult to tell. Have tried it?

cogmission:
Not yet tried it... Right now I'm in the middle of looking for adequate Linear Algebra libraries, though what I have to do is not too complex (some dot products, right vector sums) - but then I have to have a SVD implementation - [my project](https://github.com/numenta/htm.java) is reverse engineering the human neo cortex and we model columns of cells in a matrix as one of the data structures (neural nets with real neurons).

Anyway, I need to get an idea of how it might be to manipulate the bits of a Roaring Bitmap, either by extracting out to a 2D integer array doing my computations; and then putting it back in, or manipulating the bitmaps via a wrapper - in place? Just thinking about what would be most efficient...

lemire:
[...]

PR #263, opened 2018-07-06
Fast rank-peekable iterators

khitrin:
Follow-up #191/#192.

Current FastRankRoaringBitmap implementation holds a cache of containers’ cardinalities and drastically improves rank calculation in case of random access.

I propose an enhancement for sequential access case — next/advanceIfNeed followed by obtaining rank of found position.

My original case is a very sparse column implemented as bitmap and array using rank as non-empty value position.

Another example (cc #72): two sparse vectors can be implemented as bitmap-array pair and dot product of them can be calculated by intersection via advanceIfNeed and extracting values using match ranks.

I have to make a separate family of rank iterators because rank processing will reduce performance for classical usecases (except of ArrayContainerShortIterator -- getting it's sub-rank is trivial).

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
@@ -425,6 +425,12 @@ public final class ArrayContainer extends Container implements Cloneable {
   }
 
   @Override
+  public PeekableShortRankIterator getShortRankIterator() {
+    // for ArrayContainer there is no additional work, pos is known in advance
+    return new ArrayContainerShortIterator(this);
+  }
+
+  @Override
   public ContainerBatchIterator getBatchIterator() {
     return new ArrayBatchIterator(this);
   }
@@ -1295,7 +1301,7 @@ public final class ArrayContainer extends Container implements Cloneable {
 }
 
 
-final class ArrayContainerShortIterator implements PeekableShortIterator {
+final class ArrayContainerShortIterator implements PeekableShortRankIterator {
   int pos;
   ArrayContainer parent;
 
@@ -1310,11 +1316,15 @@ final class ArrayContainerShortIterator implements PeekableShortIterator {
     pos = Util.advanceUntil(parent.content, pos - 1, parent.cardinality, minval);
   }
 
+  @Override
+  public short peekNextRank() {
+    return (short) (pos + 1);
+  }
 
   @Override
-  public PeekableShortIterator clone() {
+  public PeekableShortRankIterator clone() {
     try {
-      return (PeekableShortIterator) super.clone();
+      return (PeekableShortRankIterator) super.clone();
     } catch (CloneNotSupportedException e) {
       return null;// will not happen
     }
--- a/roaringbitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
@@ -12,7 +12,7 @@ import java.nio.LongBuffer;
 import java.util.Arrays;
 import java.util.Iterator;
 
-import static java.lang.Long.numberOfLeadingZeros;
+import static java.lang.Long.bitCount;
 import static java.lang.Long.numberOfTrailingZeros;
 
 
@@ -503,6 +503,11 @@ public final class BitmapContainer extends Container implements Cloneable {
   @Override
   public PeekableShortIterator getShortIterator() {
     return new BitmapContainerShortIterator(this.bitmap);
+  }
+
+  @Override
+  public PeekableShortRankIterator getShortRankIterator() {
+    return new BitmapContainerShortRankIterator(this.bitmap);
   }
 
   @Override
@@ -1371,7 +1376,7 @@ public final class BitmapContainer extends Container implements Cloneable {
 }
 
 
-final class BitmapContainerShortIterator implements PeekableShortIterator {
+class BitmapContainerShortIterator implements PeekableShortIterator {
 
   long w;
   int x;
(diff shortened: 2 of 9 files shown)


richardstartin:
You know the position based on the number of times you've called `IntIterator.next()` so you could track the position externally. This would only work until you call `PeekableIntIterator.advanceIfNeeded()`, at which point you don't have your index into your array any more. Perhaps `PeekableIntIterator.advanceIfNeeded()` could return how many positions were skipped during the invocation as an `int`. This would work with any bitmap implementation and I would be intrigued to see how it compares to this for performance on a representative sparse vector dot product benchmark.

lemire:
At a glance, this looks fine to me.

 @richardstartin Do you have an objection regarding this code?

richardstartin:
No objection at all. I just think you could have something more general and possibly faster just by modifying advanceIfNeeded to return the number of values skipped. It’s a common use case and the space overhead of FastRankRoaringBitmap may not suit all use cases.

richardstartin:
More general in the sense it could work with RoaringBitmap and the buffer versions too.

lemire:
@khitrin  Can you add a unit test to make sure that advanceAsNeeded does what you think it does?

richardstartin:
@lemire for the avoidance of doubt, I'm not suggesting there are bugs in the `advanceIfNeeded` methods in this PR. The code looks useful and correct to me. The aim here is to yield the position of each value during iteration so that it can be used as an index into a densely packed array of positionally sparse values. This, with the ability to skip over values, enables calculations like dot products between sparse vectors.

I'm suggesting that if you have to iterate over the bitmap anyway, it's easy to track the position on the fly by counting the invocations to `IntIterator.next()`. I expect the cost of this to be negligible. The only blocker for this with existing code is that `PeekableIntIterator.advanceIfNeeded()` doesn't report how many values it skipped over. Changing this method to return the number of skipped values could generalise this work so all bitmap level iterators can yield the current position, without the need to store cumulative cardinalities. Container iterators would also not need to be modified.

lemire:
@richardstartin I also do not expect bugs: the reason I ask @khitrin to provide a unit test is just good software engineering.

Do I need to do a speech on why unit tests are a good idea?


Here are some parts of such a speech...

- Even if your original code has no bug, others might modify it. If there is no test, you won't catch the introduction a bug.
- Even if nobody modifies your code, your code might depend on code that someone modifies, thus breaking your code.

And so forth.

It is much, much easier to just always have a test.

lemire:
> Changing this method to return the number of skipped values could generalise this work so all bitmap level iterators can yield the current position, without the need to store cumulative cardinalities. Container iterators would also not need to be modified.

Might I suggest you open a new and distinct issue?

khitrin:
@lemire, thank you for this point, I will add a unit test for `PeekableIntRankIterator.advanceIfNeeded` for entire bitmap like ones for `PeekableShortRankIterator`s  ([TestRankIteratorsOfContainers.java](https://github.com/RoaringBitmap/RoaringBitmap/pull/263/files#diff-707f5fcce3b412c7740289ed945aae8e)) soon.

@richardstartin, there is some overhead for `advanceIfNeeded` in containers:
BitmapContainer: sum of bit counts over skipped words -- most suspicious for me in cases of big jumps.
RunContainer: sum of lengths over skipped runs.
ArrayContainer: no overhead.

and for `PeekableIntIterator.advanceIfNeeded` without cumulative cache: calculate and sum cardinalities over skipped containers.

Your suggestion is really good in terms of library design, but I'm afraid that some benchmarks needed over usecases not involving rank queries.

richardstartin:
> Your suggestion is really good in terms of library design, but I'm afraid that some benchmarks needed over usecases not involving rank queries.

I also advocate an empirical approach, but there are no benchmarks here to demonstrate the overhead of cardinality calculations, or to illustrate that there are actually a few dimensions to the overheads involved and there's a subtle question of how likely you are to encounter them.

For instance, if you happen to skip over an entire bitmap, there is no overhead in calculating the bitmap's cardinality: it's just a field access. You do have to perform a bit count if you skip over part of a bitmap container. But how sparse is your sparse vector if you have many bitmap containers? Sparse enough to justify the overhead of having a bitmap at all and to save space over an array?

Likewise, there will always be cardinality calculation overhead with run containers, but I doubt you'll ever have any of these in a sparse vector, otherwise the vector would likely be dense enough to better represented by a page table or possibly even a simple array, especially if you have many of them. 

So I imagine to have come up with this storage approach - and, in general, whenever the positions are more useful than the values - you probably have really sparse data, with bitmaps made up of mostly array containers, and therefore negligible cardinality calculation overhead.

My point is I think that you can easily create an API that facilitates sparse vectors in a more generic way, still useful for users who find `FastRankRoaringBitmap` too large, without cost in the expected case, but with significant cost in adversarial cases where the feature isn't actually likely to be used. Only benchmarks can justify being more specific for the sake of performance.

richardstartin:
> Might I suggest you open a new and distinct issue?

@lemire yes I should take this up elsewhere and there could be a separate implementation as descibed above.

khitrin:
I'm done with new tests.

lemire:
LGTM.

khitrin:
@lemire, @richardstartin Sorry for the disturbance, but should we try and benchmark alternative implementations or this PR could be merged as is?

richardstartin:
@khitrin that's up to @lemire, but, for what it's worth, my view is that it's a useful feature, the code is well written, but it only works with a special case subclass which:

* Is not integrated with classes like `FastAggregation`.
* Is optimised for random access rank at the cost of using extra memory.
* Penalises mutation. 

I don't think these questions have answers yet:

* Does it really need to be this restrictive for the sake of performance? 
* Do optimisations made for random access rank actually translate to sequential access? 
* If the feature were to be backported to `RoaringBitmap` and `ImmutableRoaringBitmap`, is the presented interface the right one?

lemire:
@khitrin If you recommend merging this now, it is fine with me.

khitrin:
@lemire I think it can be merged as is to enhance feature set of FastRankRoaringBitmap.

It is not a blocker for more generic approaches because a new feature will be added to already existing special subclass.

lemire:
Let us go!

PR #265, opened 2018-08-01
RoaringBitmap#bitmapOf for Integer[] and List<Integer> args

DZRock:
overload RoaringBitmap#bitmapOf for Integer[] and List<Integer> args

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -11,7 +11,9 @@ import java.io.IOException;
 import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Iterator;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


lemire:
Please make sure the PR passes our tests...
https://github.com/RoaringBitmap/RoaringBitmap/blob/master/.travis.yml

lemire:
Can you check the CI failures?

lemire:
I think your PR fails our style checks, can you correct?

```
[WARNING] src/main/java/org/roaringbitmap/RoaringBitmap.java:[511] (javadoc) SingleLineJavadoc: Javadoc comment at column 84 has parse error. Missed HTML close tag 'Integer'. Sometimes it means that close tag missed for one of previous tags.
```

Here are our instructions:

> Contributions are invited. We enforce the Google Java style. Please run mvn checkstyle:check on your code in the roaringbitmap subdirectory (module) before submitting a patch.

We do this to avoid having to debate how the code is going to be formatted.

lemire:
Ping!

lemire:
I am going to drop this since contributor is not responding.

Issue #266, opened 2018-08-07
getLongSizeInBytes() implementation for Roaring64NavigableMap


arpang:
Hi,

The getLongSizeInBytes() method of Roaring64NavigableMap class is marked as "Todo". Can someone please add the implementation or tell me the logic so that I can implement and raise a pull request?

lemire:
I think it is meant to provide an estimate of the memory usage.

PR invited.

lemire:
Closing: https://github.com/RoaringBitmap/RoaringBitmap/pull/267

PR #267, opened 2018-08-08
Introduce implemententation for Roaring64NavigableMap.getLongSizeInBy…

blacelle:
https://github.com/RoaringBitmap/RoaringBitmap/issues/266

This method was not initially implemented because I was unconfortable with estimating the usage of a TreeMap. Here is an implementation which sounds reasonnable.


richardstartin:
Where do you account for `sortedCumulatedCardinality` and `sortedHighs`? 

Note that since `FastRankRoaringBitmap` does not add the size of its cumulative cardinality cache to the size estimate, the result would be misleading whenever it's used.

lemire:
Great.

arpang:
Thanks for the quick fix @blacelle!

Issue #269, opened 2018-08-22
A better OrderedWriter for building sparse bitmaps


richardstartin:
It's quicker to use `OrderedWriter` for building bitmaps from ordered streams than repeatedly calling `RoaringBitmap.add` because it avoids the cost of locating the container and any binary searches. But using an 8KB bitset isn't the best decision when a very sparse bitmap (no bitmap containers) is to be built. Profiling shows that most time is spent in constructing array containers from bitmap containers. I think it would be quicker to have a variant of `OrderedWriter` which appends values to an `ArrayContainer` with a preallocated 4096 element `short[]`, or maybe an estimated capacity, which flushes when each 16 bit boundary is crossed.

lemire:
+1

PR #270, opened 2018-08-23
[WIP] Make OrderedWriter work better with sparse RoaringBitmaps

rafael-telles:
Solves #269 by making OrderedWriter work with both types of data (dense or sparse)

This PR still needs benchmarks.


lemire:
@rafael-telles Can you get back to @richardstartin ?

rafael-telles:
@richardstartin , I considered this too, but I decided not to have 2 classes because the user would need know which class to used based on his usage.
Maybe we can have 3 OrderedWriters: SparseOrderedWriter(what you are purposing) , DenseOrderedWriter(the original) and AdaptiveOrderedWriter(what I implemented).

lemire:
@richardstartin What do you think?

richardstartin:
I had the idea for this feature using the library to build a mask for document start positions in a binary file archive, while this isn't RoaringBitmap's sweet spot, I needed something quick, reasonably concise, that could be iterated over in both directions, to mark the positions in a mapped file.  I saw that building array containers dominated the profile, but the documents were *always* at least 512B. So in this case, I would absolutely *know* that I couldn't have more than 4096 elements per container, and the idea behind this feature was to be able to exploit that precise situation: 

- it wouldn't support out of order (in the lower 16 bits) writes (`OrderedWriter` actually does, and this is exploited in `RoaringBitmap.bitmapOfUnordered` to allow use of only a _partial_ radix sort on the higher order bits prior to building.) 
- it would just throw an exception if you get your calculations wrong etc. 

For a one-size-fits-all implementation I think `OrderedWriter` is good as it stands, and I didn't anticipate the need to change or penalise any use of it. So I prefer the idea of two separate implementations because:

1. I expect this implementation to cause performance regressions
2. I think this implementation takes up too much memory for its use case
3. I expect this implementation to break `RoaringBitmap.bitmapOfUnordered`
4. I thought it was reasonable to have two implementations: one that always works, and another that is fast when you are 100% certain your data is sparse

Is it possible to reconsider the decision not to provide two sibling implementations?

lemire:
@richardstartin Do we have a way to measure an eventual performance regression at this time?

richardstartin:
There are benchmarks for this. I can run them on my laptop to give a rough estimate.

lemire:
@richardstartin If you could give instructions, maybe @rafael-telles could run the benchmarks himself and get an idea about the performance issues that you anticipate.

I think it is always better to reason about hard numbers rather than vague estimations.

richardstartin:
There is a benchmark called `WriteSequential.incrementalUseOrderedWriter` which times how long it takes to build bitmaps for a range of different densities. You can run it from `RoaringBitmap/jmh` with 

```bash ./run.sh WriteSequential.incrementalUseOrderedWriter```

I ran it on master and got these results:

```
Benchmark                                    (randomness)    (size)  Mode  Cnt      Score       Error  Units
WriteSequential.incrementalUseOrderedWriter           0.1     10000  avgt    5     19.971 ±     0.292  us/op
WriteSequential.incrementalUseOrderedWriter           0.1    100000  avgt    5    264.715 ±     6.455  us/op
WriteSequential.incrementalUseOrderedWriter           0.1   1000000  avgt    5   2679.880 ±   136.150  us/op
WriteSequential.incrementalUseOrderedWriter           0.1  10000000  avgt    5  26165.874 ± 11343.465  us/op
WriteSequential.incrementalUseOrderedWriter           0.5     10000  avgt    5     19.697 ±     0.495  us/op
WriteSequential.incrementalUseOrderedWriter           0.5    100000  avgt    5    272.462 ±    49.170  us/op
WriteSequential.incrementalUseOrderedWriter           0.5   1000000  avgt    5   3032.112 ±   850.442  us/op
WriteSequential.incrementalUseOrderedWriter           0.5  10000000  avgt    5  25669.909 ±  1425.179  us/op
WriteSequential.incrementalUseOrderedWriter           0.9     10000  avgt    5     20.371 ±     1.350  us/op
WriteSequential.incrementalUseOrderedWriter           0.9    100000  avgt    5    272.265 ±    29.171  us/op
WriteSequential.incrementalUseOrderedWriter           0.9   1000000  avgt    5   2676.791 ±   127.749  us/op
WriteSequential.incrementalUseOrderedWriter           0.9  10000000  avgt    5  24693.047 ±  1438.060  us/op
```


And on your branch I saw:

```
Benchmark                                    (randomness)    (size)  Mode  Cnt      Score      Error  Units
WriteSequential.incrementalUseOrderedWriter           0.1     10000  avgt    5     28.352 ±    1.057  us/op
WriteSequential.incrementalUseOrderedWriter           0.1    100000  avgt    5    333.004 ±   17.180  us/op
WriteSequential.incrementalUseOrderedWriter           0.1   1000000  avgt    5   3342.633 ±  215.375  us/op
WriteSequential.incrementalUseOrderedWriter           0.1  10000000  avgt    5  33244.195 ± 1944.698  us/op
WriteSequential.incrementalUseOrderedWriter           0.5     10000  avgt    5     28.253 ±    1.161  us/op
WriteSequential.incrementalUseOrderedWriter           0.5    100000  avgt    5    331.922 ±    6.457  us/op
WriteSequential.incrementalUseOrderedWriter           0.5   1000000  avgt    5   3347.846 ±  196.953  us/op
WriteSequential.incrementalUseOrderedWriter           0.5  10000000  avgt    5  33982.642 ± 3035.831  us/op
WriteSequential.incrementalUseOrderedWriter           0.9     10000  avgt    5     28.461 ±    0.528  us/op
WriteSequential.incrementalUseOrderedWriter           0.9    100000  avgt    5    332.411 ±   14.606  us/op
WriteSequential.incrementalUseOrderedWriter           0.9   1000000  avgt    5   3306.950 ±  242.813  us/op
WriteSequential.incrementalUseOrderedWriter           0.9  10000000  avgt    5  33292.460 ± 1312.226  us/op

```

So there's a good 10-25% regression as things stand. As you work on this it would be good to track those numbers.

Also, keep in mind that this PR would break `RoaringBitmap.bitmapOfUnordered`:

```
Failed tests: 
  UnorderedWriterRandomisedTest.bitmapOfUnorderedShouldBuildSameBitmapAsBitmapOf:48 expected:<{1,2,3,4,56,103,104,105,193,265}> but was:<{2,3,1,56,4,193,105,265,104,103}>
  UnorderedWriterRandomisedTest.bitmapOfUnorderedShouldBuildSameBitmapAsBitmapOf:48 expected:<{83,84,147,167,174,267,312,313,324,421,514,554,600,601,602,622,717,718,736,737,738,833,846,909,910,911,998,1028,1054,1055,1056,1057,1058,1059,1067,1165,1234,1242,1243,1244,1257,1298,1299,1338,1339,1340,1341,1342,1360,1364,1415,1416,1417,1418,1488,1497,1498,1499,1583,1675,1710,1711,1782,1783,1784,1850,1932,2023,2024,2025,2026,2088,2089,2174,2175,2190,2191,2192,2253,2266,2267,2268,2336,2344,2440,2441,2442,2499,2500,2578,2645,2646,2734,2782,2829,2840,2921,2956,3025,3086}> but was:<{2336,1054,601,2191,1417,3025,2442,1257,1415,1418,2440,1675,1782,1165,909,2646,738,910,600,267,1360,2190,1488,1416,1244,1299,2253,1583,2192,2499,2175,2500,1058,2026,83,2782,1497,1234,602,1028,1056,2266,911,2441,84,324,1342,737,1341,2921,2344,1784,2645,2268,998,1338,1298,1340,1059,1055,1339,174,1498,1057,3086,2829,2734,1932,1242,1364,2025,2174,1711,2578,1499,2267,514,736,2023,2024,622,1067,717,554,833,2956,421,167,2840,313,147,1783,2088,1243,718,312,846,1850,1710,2089}>
  UnorderedWriterRandomisedTest.bitmapOfUnorderedShouldBuildSameBitmapAsBitmapOf:48...
```

**Comment by lemire** [comment]
Thanks @richardstartin, very useful.

**Comment by rafael-telles** [comment]
Thanks for the feedback @richardstartin , I will probably be working on it by tomorrow and will share the benchmark results with you guys.

**Comment by richardstartin** [comment]
Looks good. What shall we do when the sparse limit is exceeded? I would be inclined to throw an exception, but it might be more popular to be adaptive. With another layer of abstraction I could imagine something like this:

```java

public class SparseOrderedWriter implements OrderedWriter {
...
   DenseOrderedWriter transfer() {
   ...
   }
...
}

public class AdaptiveOrderedWriter implements OrderedWriter {
   SparseOrderedWriter sparseWriter = new SparseOrderedWriter();
   DenseOrderedWriter denseWriter = null;
   boolean sparse = true;

   @Override
   public void add(int value) {
    if (sparse) {
       try {
          sparseWriter.add(value);
      } catch (IllegalStateException e) { // or maybe return a boolean to indicate whether it was possible to add the value or not
         denseWriter = sparseWriter.transfer();
         denseWriter.add(value);
         sparse = false;
         sparseWriter = null;
     } 
    } else {
      denseWriter.add(value);
   }
}
```

So if you don't know your data very well, you use a dense writer (i.e. leave that part the way you have it now). But if as a user you expect your data to be sparse, but don't want to risk unexpected failure, you can have one of these. Profiling JIT compilers with speculative optimisations make this cost free so long as the spill never happens. You get slower but correct code if the values spill, without any weird surprises in production.

Just thinking out aloud. @lemire do you have an opinion on this?

lemire:
@richardstartin 

> What shall we do when the sparse limit is exceeded?

The current code looks reasonable, but it might be nice, at a minimum to document the issue a bit: maybe explain with a extra sentence or two how the limit is supposed to be used.

rafael-telles:
@richardstartin @lemire , I implemented the purposed AdaptiveOrderedWriter and got these benchmark results:

```
Benchmark                                    (randomness)    (size)  (writerType)  Mode  Cnt      Score      Error  Units
WriteSequential.incrementalUseOrderedWriter           0.1       100         DENSE  avgt    5      2.553 ±    0.060  us/op
WriteSequential.incrementalUseOrderedWriter           0.1       100        SPARSE  avgt    5      0.765 ±    0.010  us/op
WriteSequential.incrementalUseOrderedWriter           0.1       100      ADAPTIVE  avgt    5      0.697 ±    0.013  us/op
WriteSequential.incrementalUseOrderedWriter           0.1       100   UNOPTIMIZED  avgt    5      1.189 ±    0.010  us/op
WriteSequential.incrementalUseOrderedWriter           0.1      1000         DENSE  avgt    5      4.854 ±    0.034  us/op
WriteSequential.incrementalUseOrderedWriter           0.1      1000        SPARSE  avgt    5      2.735 ±    0.037  us/op
WriteSequential.incrementalUseOrderedWriter           0.1      1000      ADAPTIVE  avgt    5      2.682 ±    0.030  us/op
WriteSequential.incrementalUseOrderedWriter           0.1      1000   UNOPTIMIZED  avgt    5     12.481 ±    0.247  us/op
WriteSequential.incrementalUseOrderedWriter           0.1     10000         DENSE  avgt    5     18.929 ±    0.509  us/op
WriteSequential.incrementalUseOrderedWriter           0.1     10000      ADAPTIVE  avgt    5     49.886 ±    0.729  us/op
WriteSequential.incrementalUseOrderedWriter           0.1     10000   UNOPTIMIZED  avgt    5    116.078 ±    3.623  us/op
WriteSequential.incrementalUseOrderedWriter           0.1    100000         DENSE  avgt    5    260.877 ±    3.639  us/op
WriteSequential.incrementalUseOrderedWriter           0.1    100000      ADAPTIVE  avgt    5    314.580 ±    7.360  us/op
WriteSequential.incrementalUseOrderedWriter           0.1    100000   UNOPTIMIZED  avgt    5    895.785 ±   11.013  us/op
WriteSequential.incrementalUseOrderedWriter           0.1   1000000         DENSE  avgt    5   2538.726 ±   71.150  us/op
WriteSequential.incrementalUseOrderedWriter           0.1   1000000      ADAPTIVE  avgt    5   3009.198 ±   80.885  us/op
WriteSequential.incrementalUseOrderedWriter           0.1   1000000   UNOPTIMIZED  avgt    5   8920.691 ±   45.892  us/op
WriteSequential.incrementalUseOrderedWriter           0.1  10000000         DENSE  avgt    5  23483.381 ±  396.549  us/op
WriteSequential.incrementalUseOrderedWriter           0.1  10000000      ADAPTIVE  avgt    5  29962.898 ±  501.423  us/op
WriteSequential.incrementalUseOrderedWriter           0.1  10000000   UNOPTIMIZED  avgt    5  89271.744 ± 3396.370  us/op
WriteSequential.incrementalUseOrderedWriter           0.5       100         DENSE  avgt    5      2.528 ±    0.039  us/op
WriteSequential.incrementalUseOrderedWriter           0.5       100        SPARSE  avgt    5      0.763 ±    0.011  us/op
WriteSequential.incrementalUseOrderedWriter           0.5       100      ADAPTIVE  avgt    5      0.693 ±    0.104  us/op
WriteSequential.incrementalUseOrderedWriter           0.5       100   UNOPTIMIZED  avgt    5      1.229 ±    0.023  us/op
WriteSequential.incrementalUseOrderedWriter           0.5      1000         DENSE  avgt    5      5.186 ±    0.060  us/op
WriteSequential.incrementalUseOrderedWriter           0.5      1000        SPARSE  avgt    5      2.657 ±    0.018  us/op
WriteSequential.incrementalUseOrderedWriter           0.5      1000      ADAPTIVE  avgt    5      2.694 ±    0.084  us/op
WriteSequential.incrementalUseOrderedWriter           0.5      1000   UNOPTIMIZED  avgt    5     12.416 ±    0.241  us/op
WriteSequential.incrementalUseOrderedWriter           0.5     10000         DENSE  avgt    5     18.725 ±    0.231  us/op
WriteSequential.incrementalUseOrderedWriter           0.5     10000      ADAPTIVE  avgt    5     50.881 ±    9.495  us/op
WriteSequential.incrementalUseOrderedWriter           0.5     10000   UNOPTIMIZED  avgt    5    115.826 ±   28.241  us/op
WriteSequential.incrementalUseOrderedWriter           0.5    100000         DENSE  avgt    5    259.951 ±    3.922  us/op
WriteSequential.incrementalUseOrderedWriter           0.5    100000      ADAPTIVE  avgt    5    315.011 ±   15.309  us/op
WriteSequential.incrementalUseOrderedWriter           0.5    100000   UNOPTIMIZED  avgt    5    901.977 ±   25.210  us/op
WriteSequential.incrementalUseOrderedWriter           0.5   1000000         DENSE  avgt    5   2566.468 ±  163.691  us/op
WriteSequential.incrementalUseOrderedWriter           0.5   1000000      ADAPTIVE  avgt    5   3026.714 ±  103.798  us/op
WriteSequential.incrementalUseOrderedWriter           0.5   1000000   UNOPTIMIZED  avgt    5   8960.705 ±  139.529  us/op
WriteSequential.incrementalUseOrderedWriter           0.5  10000000         DENSE  avgt    5  23499.369 ±  865.911  us/op
WriteSequential.incrementalUseOrderedWriter           0.5  10000000      ADAPTIVE  avgt    5  29992.934 ±  577.696  us/op
WriteSequential.incrementalUseOrderedWriter           0.5  10000000   UNOPTIMIZED  avgt    5  89631.385 ± 2808.922  us/op
WriteSequential.incrementalUseOrderedWriter           0.9       100         DENSE  avgt    5      2.526 ±    0.028  us/op
WriteSequential.incrementalUseOrderedWriter           0.9       100        SPARSE  avgt    5      0.707 ±    0.011  us/op
WriteSequential.incrementalUseOrderedWriter           0.9       100      ADAPTIVE  avgt    5      0.687 ±    0.019  us/op
WriteSequential.incrementalUseOrderedWriter           0.9       100   UNOPTIMIZED  avgt    5      1.316 ±    0.020  us/op
WriteSequential.incrementalUseOrderedWriter           0.9      1000         DENSE  avgt    5      4.916 ±    0.071  us/op
WriteSequential.incrementalUseOrderedWriter           0.9      1000        SPARSE  avgt    5      2.670 ±    0.024  us/op
WriteSequential.incrementalUseOrderedWriter           0.9      1000      ADAPTIVE  avgt    5      2.700 ±    0.067  us/op
WriteSequential.incrementalUseOrderedWriter           0.9      1000   UNOPTIMIZED  avgt    5     12.426 ±    0.217  us/op
WriteSequential.incrementalUseOrderedWriter           0.9     10000         DENSE  avgt    5     18.797 ±    0.299  us/op
WriteSequential.incrementalUseOrderedWriter           0.9     10000      ADAPTIVE  avgt    5     50.241 ±    1.039  us/op
WriteSequential.incrementalUseOrderedWriter           0.9     10000   UNOPTIMIZED  avgt    5    106.851 ±    2.418  us/op
WriteSequential.incrementalUseOrderedWriter           0.9    100000         DENSE  avgt    5    260.020 ±    6.968  us/op
WriteSequential.incrementalUseOrderedWriter           0.9    100000      ADAPTIVE  avgt    5    313.778 ±    6.296  us/op
WriteSequential.incrementalUseOrderedWriter           0.9    100000   UNOPTIMIZED  avgt    5    665.057 ±    4.711  us/op
WriteSequential.incrementalUseOrderedWriter           0.9   1000000         DENSE  avgt    5   2532.362 ±   49.249  us/op
WriteSequential.incrementalUseOrderedWriter           0.9   1000000      ADAPTIVE  avgt    5   3035.386 ±   37.467  us/op
WriteSequential.incrementalUseOrderedWriter           0.9   1000000   UNOPTIMIZED  avgt    5   6749.062 ± 1367.867  us/op
WriteSequential.incrementalUseOrderedWriter           0.9  10000000         DENSE  avgt    5  23458.022 ±  968.460  us/op
WriteSequential.incrementalUseOrderedWriter           0.9  10000000      ADAPTIVE  avgt    5  29932.496 ± 1821.584  us/op
WriteSequential.incrementalUseOrderedWriter           0.9  10000000   UNOPTIMIZED  avgt    5  90090.239 ± 5463.192  us/op
```

Observations:

* The current AdaptiveOrderedWriter implementation is as fast as SparseOrderedWriter when it doesn't exceed the limit, as expected
* The current AdaptiveOrderedWriter is a bit slower than DenseOrderedWriter when it exceeds the limit, as expected
* There are parameters that doesn't work for SparseOrderedWriter (because of the limit), there are not showed on the benchmark output

rafael-telles:
I think it would be good if we could switch back from Dense to Sparse after flushing the DenseOrderedWriter, so that it would handle better sparse data with a few containers that exceed the limit.
I tried to do this on AdaptiveOrderedWriter:30, but realized now it makes no sense the way I tried.

roaringbitmap/src/main/java/org/roaringbitmap/SparseOrderedWriter.java
```diff
@@ -0,0 +1,175 @@
...
+   * It might receive a DenseOrderedWriter instance in order to populate it.
+   * NOTE: This method clears the current SparseOrderedWriter.
+   *
+   * @param denseWriter a nullable DenseOrderedWriter reference.
+   *                    A new DenseOrderedWriter will be created if this is null.
+   * @return a DenseOrderedWriter instance containing all non-flushed data that was
+   *         present in this SparseOrderedWriter.
+   */
+  public DenseOrderedWriter transfer(DenseOrderedWriter denseWriter) {
+    if (denseWriter == null) {
```
richardstartin: I don't follow this method. Shouldn't it just create a new DenseWriter? I don't see a case for reusing an existing instance and once the spill has occurred, I see no evidence that going back to a SparseWriter is helpful.

roaringbitmap/src/main/java/org/roaringbitmap/SparseOrderedWriter.java
```diff
@@ -0,0 +1,175 @@
...
+    if (valueIndex > 0 && Util.compareUnsigned(low, values[valueIndex - 1]) <= 0) {
+      throw new IllegalStateException("Must write in ascending order.");
+    }
+    if (valueIndex >= values.length) {
+      valuesCount = values.length;
+      throw new IllegalStateException(
+          String.format("Exceeded limit of %d values per container.", values.length));
+    }
+
+    currentKey = key;
```
richardstartin: move this into the block where the flush occurs to make it more obvious when this value can change

roaringbitmap/src/main/java/org/roaringbitmap/DenseOrderedWriter.java
```diff
@@ -0,0 +1,128 @@
+package org.roaringbitmap;
```
richardstartin: It's slightly unfortunate that this looks like new code in the diff.

roaringbitmap/src/main/java/org/roaringbitmap/AdaptiveOrderedWriter.java
```diff
@@ -0,0 +1,59 @@
...
+    this.sparse = true;
+  }
+
+  @Override
+  public void add(int value) {
+    if (sparse) {
+      try {
+        sparseWriter.add(value);
+      } catch (IllegalStateException e) {
+        denseWriter = sparseWriter.transfer(denseWriter);
```
richardstartin: This looks a bit awkward.

roaringbitmap/src/main/java/org/roaringbitmap/AdaptiveOrderedWriter.java
```diff
@@ -0,0 +1,59 @@
...
+        denseWriter.add(value);
+        sparse = false;
+      }
+    } else {
+      denseWriter.add(value);
+      if (!denseWriter.isDirty()) {
```
richardstartin: I don't see any evidence that switching back to sparse is a good idea, and probably makes the code slower.

richardstartin:
- **Heuristics**. I would be tempted to not transfer back to sparse in the adaptive version, and expect it to be faster to use a heuristic which treats a bitmap as sparse until proven otherwise, and once it is known that any of it is dense, expect to see more dense sections, and definitely don't expect the next one to be sparse. In dense cases, the adaptive writer should be only slightly behind the dense writer because it should figure out quite quickly that the data is dense.

- **Interface stability**. The existing class isn't being used by any open source users according to github search, so breaking the interface isn't a bad thing, it is used in at least one closed source code base that I am willing to update, but is it worth putting some of this stuff in an experimental package, since the interface is already being broken, and it's going to be quite complicated for someone to figure which of the many writers to use? A simplified interface appeals to me.

lemire:
Note that we have a unit test failure:

>   TestOrderedWriter.shouldThrowWhenAddingInReverseOrder[2] Expected exception: java.lang.IllegalStateException

roaringbitmap/src/main/java/org/roaringbitmap/DenseOrderedWriter.java
rafael-telles: Yes... I actually don't know how to fix this.

rafael-telles:
@richardstartin , fixed most of what you pointed out.

richardstartin:
Looks OK to me. If there are bugs, it'll be when this starts getting used that they'll really come out.

lemire:
Based on @richardstartin 's approval, let us merge this! And issue a release.

PR #271, opened 2018-08-23
Add RoaringBitmap.add with offset and length arguments

rafael-telles:
Adds `RoaringBitmap.addN` method, to allow adding multiple values to RoaringBitmap while respecting some boundaries.

This is useful when the user wants to reuse large preallocated `int[]` instances to create multiple RoaringBitmaps. Without it, the user needs to create array copies: `roaringBitmap.add(Arrays.copyOf(arr, len))`

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -446,7 +446,7 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
    * @param dat set values
    */
   public void add(final int... dat) {
-    this.add(dat, 0, dat.length);
+    this.addN(dat, 0, dat.length);
   }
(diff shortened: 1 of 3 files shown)


rafael-telles:
@lemire please take a look

lemire:
@rafael-telles  Saw it, saw it. ;-)

lemire:
Looks fine to me.

To preserve the symmetry of our API would you consider making the same change to MutableRoaringBitmap? We have duplicated code, for performance reasons... and we have to keep them in sync.

If you won't do it, I'll merge and then do the change myself.

Thanks for letting me know.

lemire:
Ping!

rafael-telles:
I will do the changes @lemire . 
Thanks!

lemire:
Looks good.

lemire:
See https://github.com/RoaringBitmap/RoaringBitmap/blob/master/AUTHORS

PR #275, opened 2018-10-18
refine and allow writing behind in ordered writer

richardstartin:
This refines `OrderedWriter` by

- making the sparse version build a `Container` as it advances, which makes it automatically "adaptive" and in the sparse case this should have no cost. 
- removing `AdaptiveOrderedWriter`
- allowing values to be added behind the current container, rather than throwing. Writing in order is advantageous, but not mandatory.
- adding getUnderlying to the `OrderedWriter` interface

I am unaware of these classes being used outside of a project I am working on, and have previously permitted @rafael-telles to make interface changes which I picked up today. Without this context, I would be less willing to make breaking changes.


lemire:
Lots good. I’ll review some more soon.

lemire:
Merging.

lemire:
There will be a new release.

richardstartin:
Thanks.

lemire:
I'm trying to issue a release. I am getting a VM crash. More later.

lemire:
It is quite ridiculous as now I am having a super hard time uploading the 300kB jar file.

lemire:
I have created several releases... using exactly the same command... I think that all of them but the last one failed to validate on the Nexus Repository. This Java infrastructure is getting more and more fragile as time passes. Hopefully, it is ok now.

PR #279, opened 2018-10-27
remove optimiseForBitmaps because it's usually a bad idea

richardstartin:
I spoke to @ppiotrow and we agree that the optimise for bitmaps option was a bad idea, and would require changes to the container implementations to do the right thing all the time, so this PR removes it.


richardstartin:
Travis seems to be quite creaky these days.

lemire:
Restarting travis job....

lemire:
We are now good.

PR #284, opened 2018-11-12
implement previousValue, port nextValue and previousValue to buffer p…

richardstartin:
…ackage


richardstartin:
Looks like we have a flaky test:

```Failed tests: 
  BitmapAnalyserTest.analyseRandomBitmap:39 expected:<0.2> but was:<0.0>```

When I put the test in an infinite loop it breaks within 10 seconds, because it depends on the RNG producing a number < 0.2 within 80 invocations which is probable but not guaranteed.

richardstartin:
There are a few things going on in the flaky test, which aren’t really the test’s fault.

- Random choice of container type within randomBitmap
- data created within randomBitmap ends up as unexpected container type. Bitmaps get run compressed, some runs get turned into arrays and vice versa.
- randomBitmap’s output wasn’t ever expected to be tested but is usually reasonably close to the supplied parameters 

It would be interesting to fuzz test whether statistics are preserved during bitmap -> int[] -> bitmap conversions.

lemire:
@richardstartin 

I have created a branch with deterministic tests. I think we should move to deterministic tests.

I won't merge this branch now because I fear how it might interact with your work. Can you have a look?

https://github.com/RoaringBitmap/RoaringBitmap/tree/makingtestsdeterministics

richardstartin:
I agree but think it could be made deterministic with a lighter touch. I will add a commit here, let me know what you think.

lemire:
@richardstartin  That's fine. I am not in love with my version. 

I think that having non-deterministic tests is a super bad idea. I feel a bit of shame that I did not catch that when it was committed. :-(

richardstartin:
I introduced them with the fuzz tests, and I think they make sense when the nondeterminism is independent of the assertions made. They can help verify statements like "property X should hold for all possible inputs" avoiding the (impossible) construction of all inputs. Each build being an independent event makes the probability of the existence of a bug given N sequential passed builds tend to zero quickly. It was in this spirit that the changes were introduced, If you make assertions _about_ the randomness itself (such as "within 1000 trials the RNG converges to a uniform distribution") you are in a lot of trouble.

Please take a look at the latest commit which makes the tests deterministic.

lemire:
Looks good to me. Let us see what travis thinks.

richardstartin:
Of course, it passes, because the seed was chosen to accommodate the failing test. But all the tests making assertions independent of the statistics of the random number generator are now a lot weaker because they run exactly the same way each time, and if there are undiscovered bugs they will not find them, because the seed has been fixed to make them pass.

lemire:
@richardstartin 

I think that the Go version of roaring go fuzzy tests first. The way it works over there is that we have standard, deterministic tests... then we have a separate fuzzy engine that you can run for extended periods of time (indefinitely in fact). I think that's a good design.

Once the fuzzy tests detect a problem, this problem can then become a deterministic test to avoid regression.

It is quite fun. Take an idle machine, then start the tests... let it run for days... (I did that!)... see if it finds something!


In any case, do you recommend merging your code?

richardstartin:
I would like to find a way to keep the randomness of the property based tests, since if you have a test for a condition which should hold regardless of its input, you don't need to fix the input and can only gain from the ability to generate random test cases. It's important to distinguish between this style of testing and tests which have constraints on inputs with nondeterministic input. So I'm not keen to hastily neuter the fuzz tests, so I'm not in a rush to merge this or have it merged at all for that matter. 

Is there a way to address your concerns about nondeterminism while keeping the tests? What about running them in a different module or in a different stage of the build?

lemire:
Ok. Let us discuss the objectives.

What I'd like is for the CI tests to be deterministic. Either they always fail or they always pass. If I restart a test, it should come out the same, no matter what. You should be able, on your machine, to reproduce exactly (up to memory failures and timeouts) the CI results.

This should also be true during release. Whether the tests pass during a release should be pre-determined. E.g., you don't want me to try to release, get an error, have me try again, and get the release to pass. 

This does not preclude non-determinism... but that should be an option.

Do we agree?

lemire:
Ok. Let us delay this.

richardstartin:
@lemire I've moved the fuzz tests into a dedicated module which will only run in the fuzz-test profile. This will likely decrease code coverage noticeably, but makes the CI build deterministic. I'm happy for you to merge this PR if you are.

lemire:
Let us go then. Merging.

PR #288, opened 2018-12-13
Fix Java 8 compatibility issue

clintropolis:
Fixes issue using artifacts built with jdk 9+ when run on java 8, which produces errors of the form:

```
java.lang.NoSuchMethodError: java.nio.ByteBuffer.limit(I)Ljava/nio/ByteBuffer;
	at org.roaringbitmap.buffer.ImmutableRoaringArray.<init>(ImmutableRoaringArray.java:51)
	at org.roaringbitmap.buffer.ImmutableRoaringBitmap.<init>(ImmutableRoaringBitmap.java:968)
...
```

This is due to changes to several `ByteBuffer` methods in java 9+ resulting in bytecode generated that includes mismatched return types when built with jdk 9 or newer, [as described in this issue](https://jira.mongodb.org/browse/JAVA-2559).

This PR fixes this issue by adding the `release` compiler flag, described in [JEP 247](http://openjdk.java.net/jeps/247). This is done by adding an additional maven profile that activates if jdk version used to build is 9 or greater which sets the property `maven.compiler.release` adding `--release=8` to the compiler flags.The `allModules` profile activation has been changed to be always applied unless the `release` profile is in use, since `activeByDefault` only applies a profile if no others are active.

### DIFF
--- a/fuzz-tests/pom.xml
+++ b/fuzz-tests/pom.xml
@@ -43,8 +43,7 @@
                         <version>3.6.0</version>
                         <configuration>
                             <compilerVersion>${javac.target}</compilerVersion>
-                            <source>${javac.target}</source>
-                            <target>${javac.target}</target>
+                            <release>${javac.release}</release>
                             <verbose>true</verbose>
                             <fork>true</fork>
                             <showDeprecation>true</showDeprecation>
--- a/fuzz-tests/pom.xml
+++ b/fuzz-tests/pom.xml
@@ -43,7 +43,8 @@
                         <version>3.6.0</version>
                         <configuration>
                             <compilerVersion>${javac.target}</compilerVersion>
-                            <release>${javac.release}</release>
+                            <source>${javac.target}</source>
+                            <target>${javac.target}</target>
                             <verbose>true</verbose>
                             <fork>true</fork>
                             <showDeprecation>true</showDeprecation>
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <javac.target>1.8</javac.target>
+        <javac.release>8</javac.release>
 
         <!-- This location is OK for main module but not sub-modules -->
         <checkstyle.configLocation>${basedir}/roaringbitmap/style/roaring_google_checks.xml</checkstyle.configLocation>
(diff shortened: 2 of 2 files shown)


clintropolis:
Heh... and I didn't confirm that it could still be built with java 8, this doesn't work as is because the flag doesn't exist there! Maybe another profile that is used to build publishing artifacts is the solution? Thoughts?

lemire:
We can drop support for java 8 in travis. Of course, while we are at it, we should maybe drop support for java 8 in upcoming versions.

Java 8 has reached its end of public updates.

lemire:
Anyhow, someone who knows maven well should review this. 

Maybe @blacelle ?

brharrington:
Updating travis to only run on jdk11 and build with `--release=8` seems like a reasonable compromise to dropping jdk8 support entirely. 

That said, you should be able to build using one jdk and run the tests using another by switching the jdk as part of the script. See:

https://docs.travis-ci.com/user/languages/java/#switching-jdks-java-8-and-below-within-one-job
https://docs.travis-ci.com/user/languages/java/#switching-jdks-to-java-10-and-up-within-one-job

I'm not sure if it is worth the extra complexity. Perhaps this could be revisited if problems are seen in practice.

clintropolis:
I would agree it's hasty to drop java 8 support. 

I don't believe that the artifacts need to be different for different java versions the release unless there are expected performance improvements to be had targeting a newer version, as it tests fine all all java greater than 8 I tried when built with these flags. 

I saw the travis docs and that would make sense as a potential avenue if we want to add tests to ensure that building with jdk 11 functions on jdk 8.

For now, I'm trying to see if I can selectively enable the `release` compiler flag automatically with a maven profile activation if the jdk is greater than 8, which seems like it could maybe be the best solution if I can get it to do the right thing in all the right places.

clintropolis:
The latest version of this patch adds an additional maven profile that activates if jdk version used to build is 9 or greater which sets the property `<maven.compiler.release>` adding `--release=8` to the compiler flags. The `allModules` profile activation has been changed to be always applied unless the `release` profile is in use, since `activeByDefault` only applies a profile if no others are active.

Lightly tested locally that artifacts built with jdk 8 or jdk 9+ work on 8+ runtimes.

lemire:
Let us try this out. We will revisit if it does not solve the problem.

lemire:
I have issued a new release.

clintropolis:
Thanks! I don't see a new version published yet, but I will test it out once it's on the repository (unless it's supposed to be there and it's absence is a casualty of the changes in this patch).

lemire:
@clintropolis For reasons that I do not fully understand, I had to partially revert your PR to get it to even try to upload the artefacts...

https://github.com/RoaringBitmap/RoaringBitmap/commit/8c1a714187abe8103e0a17896a549817872f33c7

It should be fine now.

lemire:
Please issue a PR if you think that it needs to be brought back.

PR #290, opened 2018-12-15
Mismatch intrinsic

richardstartin:
This delegates `ArrayContainer.equals(ArrayContainer)` and `RunContainer.equals(RunContainer)` to `ArraysShim` which is compiled at 8 by default, and places another version which calls into a new `Arrays` intrinsic in META-INF/versions/11, which can be loaded whenever JDK11 is in use.

Using the vectorized equals method, I saw gains of at least 5x. Weirdly, I can't verify this within the project, and can only run benchmarks to verify this from an external project. I don't fully understand why.


richardstartin:
Note that this also drops the openjdk8 travis build. This is because we produce java 8 bytecode with javac 11, but javac 8 would not be able to compile the java 11 version of ArraysShim.

richardstartin:
Jacoco doesn’t seem to like Multi-Release by default, will look at workarounds.

richardstartin:
Jacoco seems not to support multirelease jars yet. https://github.com/jacoco/jacoco/issues/407
I have excluded the byte code under META-INF from Jacoco analysis for now.

richardstartin:
Having updated the shaded jars to be multi-release, here is the performance improvement:

JDK11 (uses vectorized intrinsic)
```
Benchmark                     Mode  Cnt  Score   Error  Units
EqualsBenchmark.equalBitmaps  avgt    5  7.661 ± 0.255  us/op
```

Multi release RoaringBitmap jar built with JDK11, run with JDK8 (no vectorized instrinsic, jar still useable by JDK8 users)
```
Benchmark                     Mode  Cnt   Score   Error  Units
EqualsBenchmark.equalBitmaps  avgt    5  28.052 ± 5.472  us/op
```

lemire:
It is interesting that java won't, on its own, vectorize a comparison loop. The gains definitively appear to be worth it.

lemire:
Merging.

richardstartin:
Probably worth checking the release still works after merging

lemire:
Nothing works after merging.

lemire:
Now it works. It is fine to make `shims` into a distinct module, but it needs to be built and released. It will also become an artefact with the next release. If that's not your intention...

richardstartin:
Actually I wasn't absolutely sure how this would interact with the release. If it's necessary to release shims, then perhaps there is no good reason to shade it into RoaringBitmap itself.

lemire:
I am open to every design possibility.

Let us hold up producing new releases until we are happy.

richardstartin:
The only important thing to do is ensure that the released artifact has Multi-Release set in MANIFEST.MF. I can put in another PR quickly to replace the shading with creation of a multi release jar. I don't think releasing the shims module is such a bad thing, any more than releasing RoaringBitmapParent, but let's remove the bungled workaround (shading).

PR #291, opened 2018-12-15
Rank iterator short overflow bugfix

khitrin:
Unfortunately, I was introduced a very stupid bug in #263 for rank iterators.

Rank range is 1-65536, not 0-65535, so a short overflow happens for full containers.

There is a fix w/ regression tests.


lemire:
Great.

PR #293, opened 2018-12-18
simplify pom since jdk8 build support dropped

clintropolis:
Since #290 dropped support for building with jdk8, the first version of #288 is a simpler setup since we can drop the extra profile. Tested locally that built artifact still functions with java 8.


lemire:
Thanks.

PR #295, opened 2018-12-20
refine batch iterators 

richardstartin:
- make `BatchIterator` cloneable
- make it easily convertible to `IntIterator`


lemire:
Merging.

PR #300, opened 2018-12-21
Public shims

richardstartin:
Makes the shims module public, and it will have javadoc.


lemire:
Restarting travis job.

Issue #196, opened 2017-12-13
Implement partially mapped bitmaps: hybrid Java heap + off heap


lemire:
This is an issue raised by @kishoreg and related to how [Pinot](https://github.com/linkedin/pinot/wiki) uses Roaring bitmaps.

Currently, we support ``ImmutableRoaringBitmap``... ``MutableRoaringBitmap`` and ``RoaringBitmap``. The ``ImmutableRoaringBitmap`` class stores the containers off-heap (possibly on disk). Other classes (``MutableRoaringBitmap`` and ``RoaringBitmap``) store the data on Java's heap.

However, there are instances where we would like just **one container** to be stored in memory and to be mutable, whereas the rest of the containers should be off-heap and immutable.

I think that we are lucky in that this can be implemented in a sane manner, without having to go crazy and risk lots of bugs. However, I have not yet come up with a design that I like, so I want to open this up for discussion. Another issue is that I am not 100% clear on what would best serve Pinot's needs.

A ``MutableRoaringBitmap`` is implemented as a derived class of ``ImmutableRoaringBitmap``. 
The key implementation concept is the attribute  ``PointableRoaringArray highLowContainer`` in the class ``ImmutableRoaringBitmap``. We currently have two types of ``PointableRoaringArray``. The base ``ImmutableRoaringBitmap`` class uses an ``ImmutableRoaringArray``. Unsurprisingly, the ``MutableRoaringBitmap`` uses a ``MutableRoaringArray``.

If one looks at the ``MutableRoaringBitmap`` class itself, it is rather thin. Maybe 1500 lines, but a lot of it may not be needed in a hybrid model, except maybe for an ``add`` method. That is, we inherit many useful methods from ``ImmutableRoaringBitmap``.

I think that one possibility would be to implement something like a ``HybridRoaringBitmap`` that would be an instance of the base base ``ImmutableRoaringBitmap``. It would have an ``ImmutableRoaringArray`` but also a ``MappeableContainer``. 

It might behave much like a ``MutableRoaringBitmap``, except that any attempt at modifying a mapped container would generate an exception. I am not sure how you make sure that there is only ever just one mutable container.

I will stop here and open up the discussion.

See also https://github.com/RoaringBitmap/RoaringBitmap/issues/193 which might ever offer a viable alternative.

kishoreg:
Thanks for starting creating this issue your initial analysis of the existing code is encouraging. 

[...]

PR #302, opened 2019-01-02
support contains on RoaringBitmapWriter without flushing

richardstartin:
This enables a contains query on a bitmap being appended to without enforcing a flush, which improves subsequent insertion performance.


lemire:
I'll let the CI tests run.

richardstartin:
There are more similar changes I want to make so please don't merge this yet.

lemire:
Ok.

richardstartin:
I'm aiming to bring `RoaringBitmapWriter` under the `BitmapDataProvider` interface so a bitmap in construction can be used as a bitmap with little or no penalty. This is a similar idea to the discussion of hybrid bitmaps in #196.

PR #305, opened 2019-02-15
Implement next/previousAbsentValue

### LINKED ISSUE #282: Support previousValue, nextAbsentValue, and previousAbsentValue
Hi,

Both `BitSet` and `SparseBitSet` have said functions, which return the next / previous index which is contained / not contained in the set. All of them should be rather easy to implement I guess, since they are very similar to nextValue / (reverse) iterators.

I'm using roaring bitmaps in a [library](https://github.com/incaseoftrouble/naturals-util) abstracting data structures used to represent sets of natural numbers. Among other things, the library offers complement views on such sets, where the role of present and absent values is flipped. Naturally, such a "next/previousAbsentValue" is useful in this context.

incaseoftrouble:
Closes #282 

I think my git "fixed" the line endings to all be '\n'. View the merge with `?w=1` to ignore the whitespace changes. I'm not sure how to fix this :S.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
@@ -2316,8 +2316,11 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
   public long previousValue(int fromValue) {
     short key = Util.highbits(fromValue);
     int containerIndex = highLowContainer.advanceUntil(key, -1);
+    if (containerIndex == highLowContainer.size()) {
+      return -1L;
+    }
     long prevSetBit = -1L;
-    while (containerIndex != -1 && containerIndex < highLowContainer.size() && prevSetBit == -1L) {
+    while (containerIndex != -1 && prevSetBit == -1L) {
       short containerKey = highLowContainer.getKeyAtIndex(containerIndex);
       Container container = highLowContainer.getContainerAtIndex(containerIndex);
       int bit = (Util.compareUnsigned(containerKey, key) < 0
@@ -2329,6 +2332,92 @@ public class RoaringBitmap implements Cloneable, Serializable, Iterable<Integer>
     assert prevSetBit <= 0xFFFFFFFFL;
     assert prevSetBit <= Util.toUnsignedLong(fromValue);
     return prevSetBit;
+  }
+
+  @Override
+  public long nextAbsentValue(int fromValue) {
+    long nextAbsentBit = computeNextAbsentValue(fromValue);
+    assert nextAbsentBit <= 0xFFFFFFFFL;
+    assert nextAbsentBit >= Util.toUnsignedLong(fromValue);
+    assert !contains((int) nextAbsentBit);
+    return nextAbsentBit;
+  }
+
+  private long computeNextAbsentValue(int fromValue) {
+    short key = Util.highbits(fromValue);
+    int containerIndex = highLowContainer.advanceUntil(key, -1);
+
+    int size = highLowContainer.size();
+    if (containerIndex == size) {
+      return Util.toUnsignedLong(fromValue);
+    }
+    short containerKey = highLowContainer.getKeyAtIndex(containerIndex);
+    if (fromValue < containerKey << 16) {
+      return Util.toUnsignedLong(fromValue);
+    }
+    Container container = highLowContainer.getContainerAtIndex(containerIndex);
+    int bit = container.nextAbsentValue(Util.lowbits(fromValue));
+    while (true) {
+      if (bit != 1 << 16) {
+        return Util.toUnsignedLong((containerKey << 16) | bit);
+      }
+      assert container.last() == (1 << 16) - 1;
+      if (containerIndex == size - 1) {
+        return Util.toUnsignedLong(highLowContainer.last()) + 1;
+      }
+
+      containerIndex += 1;
+      short nextContainerKey = highLowContainer.getKeyAtIndex(containerIndex);
+      if (containerKey + 1 < nextContainerKey) {
+        return Util.toUnsignedLong((containerKey + 1) << 16);
+      }
+      containerKey = nextContainerKey;
+      container = highLowContainer.getContainerAtIndex(containerIndex);
+      bit = container.nextAbsentValue((short) 0);
+    }
+  }
+
+  @Override
+  public long previousAbsentValue(int fromValue) {
+    long prevAbsentBit = computePreviousAbsentValue(fromValue);
+    assert prevAbsentBit <= 0xFFFFFFFFL;
+    assert prevAbsentBit <= Util.toUnsignedLong(fromValue);
+    assert !contains((int) prevAbsentBit);
+    return prevAbsentBit;
+  }
+
+  private long computePreviousAbsentValue(int fromValue) {
+    short key = Util.highbits(fromValue);
+    int containerIndex = highLowContainer.advanceUntil(key, -1);
+
+    if (containerIndex == highLowContainer.size()) {
+      return Util.toUnsignedLong(fromValue);
+    }
+    short containerKey = highLowContainer.getKeyAtIndex(containerIndex);
+    if (fromValue < containerKey << 16) {
+      return Util.toUnsignedLong(fromValue);
+    }
+    Container container = highLowContainer.getContainerAtIndex(containerIndex);
+    int bit = container.previousAbsentValue(Util.lowbits(fromValue));
+
+    while (true) {
+      if (bit != -1) {
+        return Util.toUnsignedLong((containerKey << 16) | bit);
+      }
+      assert container.first() == 0;
+      if (containerIndex == 0) {
+        return Util.toUnsignedLong(highLowContainer.first()) - 1;
+      }
+
+      containerIndex -= 1;
+      short nextContainerKey = highLowContainer.getKeyAtIndex(containerIndex);
+      if (nextContainerKey < containerKey - 1) {
+        return Util.toUnsignedLong((containerKey << 16)) - 1;
+      }
+      containerKey = nextContainerKey;
+      container = highLowContainer.getContainerAtIndex(containerIndex);
+      bit = container.previousAbsentValue((short) ((1 << 16) - 1));
+    }
   }
 
   /**
--- a/roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
@@ -1269,27 +1269,26 @@ public final class MappeableBitmapContainer extends MappeableContainer implement
   }
 
   /**
-   * Find the index of the next unset bit greater or equal to i, returns -1 if none found.
+   * Find the index of the next clear bit greater or equal to i.
    *
    * @param i starting index
-   * @return index of the next unset bit
+   * @return index of the next clear bit
    */
-  public short nextUnsetBit(final int i) {
-    int x = i / 64;
+  public int nextClearBit(final int i) {
+    int x = i >> 6; // i / 64 with sign extension
     long w = ~bitmap.get(x);
     w >>>= i;
     if (w != 0) {
-      return (short) (i + numberOfTrailingZeros(w));
+      return i + numberOfTrailingZeros(w);
     }
-    ++x;
-    // for speed, replaced bitmap.limit() with hardcoded MAX_CAPACITY / 64
-    for (; x < MAX_CAPACITY / 64; ++x) {
-      long X = bitmap.get(x);
-      if (X != ~0L) {
-        return (short) (x * 64 + numberOfTrailingZeros(~X));
+    int length = bitmap.limit();
+    for (++x; x < length; ++x) {
+      long map = ~bitmap.get(x);
+      if (map != 0L) {
+        return x * 64 + numberOfTrailingZeros(map);
       }
     }
-    return -1;
+    return MAX_CAPACITY;
   }
 
   @Override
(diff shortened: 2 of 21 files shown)


incaseoftrouble:
Also I'm not sure whether I added the tests correctly. I didn't find a unified testing framework, so I more or less copypasted the code all over the place ...

lemire:
Thanks. I think we need to patch things up before we can merge this. We don't want white-space changes all over the place.

incaseoftrouble:
Fixed (I think).

Although settling for one line ending scheme would be a nice to have :)

richardstartin:
The build failure is unrelated to this PR, is new, and might be related to [this](https://bugs.openjdk.java.net/browse/JDK-8212233). I will try to fix it in another PR since I introduced the shims module.

lemire:
Line-ending consistency might be nice, indeed. I am sure we all agree that it should probably be handled by its own PR with additional style-related tests. We already enforce quite a bit regarding style, and if we don't enforce line-endings, we could/should do it as well.

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
```diff
@@ -1269,27 +1269,26 @@ public int nextSetBit(final int i) {
...
-   * Find the index of the next unset bit greater or equal to i, returns -1 if none found.
+   * Find the index of the next clear bit greater or equal to i.
    *
    * @param i starting index
-   * @return index of the next unset bit
+   * @return index of the next clear bit
    */
-  public short nextUnsetBit(final int i) {
-    int x = i / 64;
+  public int nextClearBit(final int i) {
```
richardstartin: I'm not sure if anyone is using this. I prefer the new name though.

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
```diff
@@ -1269,27 +1269,26 @@ public int nextSetBit(final int i) {
...
+    int x = i >> 6; // i / 64 with sign extension
```
richardstartin: I would prefer we don't put comments like this in... It's similar to the famous one below :)

```java
// increment i
++i;
```

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
```diff
@@ -1269,27 +1269,26 @@ public int nextSetBit(final int i) {
...
-        return (short) (x * 64 + numberOfTrailingZeros(~X));
+    int length = bitmap.limit();
+    for (++x; x < length; ++x) {
+      long map = ~bitmap.get(x);
+      if (map != 0L) {
+        return x * 64 + numberOfTrailingZeros(map);
       }
     }
-    return -1;
+    return length * 64;
```
richardstartin: The maximum possible value in the container is the constant `0xFFFF`, and I wonder if C2 would figure out that bitmap.limit() really does always return 1024 in this case.

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
richardstartin: Should this return an `int`? Was the decision to change this for efficiency, because you will typically immediately convert it back to an `int`?

roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
```diff
@@ -2316,8 +2316,11 @@ public long nextValue(int fromValue) {
   public long previousValue(int fromValue) {
     short key = Util.highbits(fromValue);
     int containerIndex = highLowContainer.advanceUntil(key, -1);
+    if (containerIndex == highLowContainer.size()) {
```
richardstartin: Good catch.

richardstartin:
I did a quick review, I only have minor cosmetic gripes. However, I ran a fuzz test, after 3 minutes I hit an assertion error.

```java
  @Test
  public void nextAbsentValueIsAlwaysAbsent() {
    verifyInvariance(rb -> !rb.isEmpty(), (i, rb) -> {
      long nextAbsentValue = rb.nextAbsentValue(rb.select(i));
      return (nextAbsentValue == -1L && i == rb.getCardinality()) || !rb.contains((int)nextAbsentValue);
    });
  }
```

```
java.lang.AssertionError
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:598)
	at java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:677)
	at java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:735)
	at java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:160)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateParallel(ForEachOps.java:174)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:418)
	at org.roaringbitmap.Fuzzer.verifyInvariance(Fuzzer.java:166)
	at org.roaringbitmap.Fuzzer.verifyInvariance(Fuzzer.java:155)
	at org.roaringbitmap.Fuzzer.nextAbsentValueIsAlwaysAbsent(Fuzzer.java:325)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.lang.AssertionError
	at org.roaringbitmap.RoaringBitmap.nextAbsentValue(RoaringBitmap.java:2363)
	at org.roaringbitmap.Fuzzer.lambda$nextAbsentValueIsAlwaysAbsent$66(Fuzzer.java:326)
	at org.roaringbitmap.Fuzzer.lambda$verifyInvariance$17(Fuzzer.java:168)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
	at java.util.stream.IntPipeline$4$1.accept(IntPipeline.java:250)
	at java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:110)
	at java.util.Spliterator$OfInt.forEachRemaining(Spliterator.java:693)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:291)
	at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
```

This trips the assert at below:

```java
  @Override
  public long nextAbsentValue(int fromValue) {
    short key = Util.highbits(fromValue);
    int containerIndex = highLowContainer.advanceUntil(key, -1);
    long nextAbsentBit = -1L;

    int size = highLowContainer.size();
    if (containerIndex == size) {
      nextAbsentBit = Util.toUnsignedLong(fromValue);
    } else {
      while (containerIndex < size && nextAbsentBit == -1L) {
        short containerKey = highLowContainer.getKeyAtIndex(containerIndex);
        Container container = highLowContainer.getContainerAtIndex(containerIndex);
        int bit = Util.compareUnsigned(containerKey, key) > 0
                ? container.nextAbsentValue((short) 0)
                : container.nextAbsentValue(Util.lowbits(fromValue));
        nextAbsentBit = bit == (1 << 16) ? -1L : Util.toUnsignedLong((containerKey << 16) | bit);
        ++containerIndex;
      }
      if (nextAbsentBit == -1L) {
        // All containers are completely filled
        nextAbsentBit = Util.toUnsignedLong(highLowContainer.last()) + 1;
      }
    }

    assert nextAbsentBit <= 0xFFFFFFFFL;
    // this one failed 
    assert nextAbsentBit >= Util.toUnsignedLong(fromValue);
    assert !contains((int) nextAbsentBit);
    return nextAbsentBit;
  }
```

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
incaseoftrouble: I slightly changed the semantics (return `highestBit() + 1` instead of `-1`), its in line with `BitSet` now.

It was unused inside the library. Could keep both maybe?

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
incaseoftrouble: True, a similar comment is present in `nextSetBit`. I can remove both.

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
incaseoftrouble: Symmetry, `nextSetBit` also returns an int.

BTW, AFAIK java lifts to ints on any operation, the only benefit short has over int is that arrays are stored denser.

incaseoftrouble:
Ah I didn't know how to run the fuzz tests. Thanks a lot, I will look into it again. I guess its a simple corner case I overlooked.

roaringbitmap/src/main/java/org/roaringbitmap/buffer/MappeableBitmapContainer.java
richardstartin: It does, but preserves sign when it does so, which would be the wrong thing to do here (and is the source of a lot of bugs in this library).

richardstartin: No I think the new name is a good idea.

incaseoftrouble:
I extended the fuzz test to cross check with java's BitSet implementation. This does not work with indices > `Integer.MAX_VALUE`, so there might be some cornercases there.

I found some more bugs which may need a bit of time to fix. I hope I have some progress in the next days.

lemire:
@incaseoftrouble 

> By mapping roaring bitmaps to the immutable counterpart, I should get coverage on all the `MappeableXY` things, right?

Can you elaborate?

lemire:
@incaseoftrouble 

I suggest you synchronize with master, we have changed the `.travis` file, effectively removing jdk11. This should allow tests to pass.

It looks like the Java modules fail our code at the javadoc generation level. This should be no more than a warning. This has been reportedly fixed in jdk12 but not in jdk11.

incaseoftrouble:
> I suggest you synchronize with master, we have changed the `.travis` file, effectively removing jdk11. This should allow tests to pass.

Thanks, will do.

> Can you elaborate?

There are all these `MappeableXYContainer` things, which are used in the `buffer` package. I also added the next/prev absent things there. I just realized there is a `BufferFuzzer` which does what I wanted :-)

incaseoftrouble:
Fixed all. Tests are going through. More comments?

PR #308, opened 2019-02-16
demonstrate javadoc bug caused by jdk11

richardstartin:


lemire:
Let us drop jdk11 for the time being.

Issue #309, opened 2019-02-19
BufferParallelAggregation throws due to faulty partitioning function


ppiotrow:
I've migrated our bitmap execution software from VM with 8 cores to cloud where I enabled cpu shares corresponding to 32 cores according to `Runtime.getRuntime.availableProcessors()`.
In the logs started appearing this error quite frequently (~1 per 10 requests fails).
I'm running on RoaringBitmap `0.7.21`

```
Caused by: java.lang.IllegalArgumentException: fromIndex(837) > toIndex(833)
	at java.util.ArrayList.subListRangeCheck(ArrayList.java:1014)
	at java.util.ArrayList.subList(ArrayList.java:1004)
	at org.roaringbitmap.buffer.BufferParallelAggregation.lambda$or$12(BufferParallelAggregation.java:229)
	at java.util.stream.IntPipeline$4$1.accept(IntPipeline.java:250)
	at java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:110)
	at java.util.Spliterator$OfInt.forEachRemaining(Spliterator.java:693)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:747)
	at java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:721)
	at java.util.stream.AbstractTask.compute(AbstractTask.java:316)
	at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool.helpComplete(ForkJoinPool.java:1870)
	at java.util.concurrent.ForkJoinPool.awaitJoin(ForkJoinPool.java:2045)
	at java.util.concurrent.ForkJoinTask.doInvoke(ForkJoinTask.java:404)
	at java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:734)
	at java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:714)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
[...]

PR #310, opened 2019-02-19
Fix bug in partitioning containers for parallel aggregation

ppiotrow:


ppiotrow:
fixes #309

lemire:
I have reviewed the math and it looks good to me.

Issue #313, opened 2019-02-26
Assertion Error in `nextValue`


incaseoftrouble:
The following code throws an assertion error

```java
    RoaringBitmap r = new RoaringBitmap();
    r.add(0, 1, 2, 4, 6);
    assertEquals(-1, r.nextValue(7));
```

With assertions disabled, the `nextValue` call returns 0. I already have a fix, I just have to convince my IDE not to make any whitespace changes ...

PR #315, opened 2019-02-26
Fix lineendings

incaseoftrouble:
I added a `.gitattributes` file as suggested by @blacelle in https://github.com/RoaringBitmap/RoaringBitmap/commit/7c1e4a398d9e107bd6d5e45eebbf813cceef0bce and ran `git add --renormalize` as suggested by the [git docs](https://git-scm.com/docs/gitattributes)


lemire:
@incaseoftrouble Can you sync with master now that the PR on ArrayContainer PR was pulled in relation with https://github.com/RoaringBitmap/RoaringBitmap/issues/313

I could do it but I think that we are less likely to encounter a problem if the same person fixes all the line endings (presumably using the same approach throughout).

incaseoftrouble:
Done. As indicated by the diff ( https://github.com/RoaringBitmap/RoaringBitmap/pull/315/files?w=1 ) there are only whitespace changes aside from the addition of `.gitattributes`

.gitattributes
```diff
@@ -0,0 +1,19 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto eol=lf
```
blacelle: I suggest leaving text=auto and not eol=lf

People working under Windows might get into trouble with this.

incaseoftrouble: Done. Wasn't sure which is better.

lemire:
I have verified that the changes are white-space only. Merging.

PR #320, opened 2019-03-15
Faster deserialization by loading BitmapContainer in a single .readFully

blacelle:
The IO effect of reading bytes one by one seems quite big. With a single .readFully:

recommended: 
247 ms 107 ms 83 ms 177 ms 81 ms 81 ms 107 ms 83 ms 83 ms 116 ms 154 ms 97 ms 82 ms 85 ms 83 ms 84 ms 97 ms 84 ms 84 ms 83 ms 83 ms 117 ms 161 ms 132 ms 82 ms 84 ms 84 ms 84 ms 84 ms 82 ms 
average: 103
via ByteArrayInputStream: 
257 ms 86 ms 86 ms 86 ms 86 ms 84 ms 82 ms 82 ms 180 ms 88 ms 83 ms 85 ms 84 ms 84 ms 85 ms 86 ms 102 ms 90 ms 84 ms 82 ms 83 ms 84 ms 83 ms 85 ms 94 ms 82 ms 81 ms 82 ms 84 ms 82 ms 
average: 94
via Immutable: 
130 ms 69 ms 77 ms 67 ms 67 ms 70 ms 67 ms 64 ms 62 ms 64 ms 77 ms 65 ms 64 ms 63 ms 64 ms 64 ms 64 ms 62 ms 66 ms 64 ms 66 ms 67 ms 64 ms 65 ms 64 ms 62 ms 78 ms 66 ms 65 ms 64 ms 
average: 68

Reference: https://github.com/RoaringBitmap/RoaringBitmap/issues/319


blacelle:
leaving little-endian management to byteBuffer:

recommended: 
256 ms 81 ms 77 ms 142 ms 75 ms 75 ms 95 ms 74 ms 75 ms 99 ms 138 ms 77 ms 73 ms 74 ms 76 ms 131 ms 75 ms 74 ms 74 ms 74 ms 74 ms 133 ms 170 ms 74 ms 76 ms 80 ms 77 ms 77 ms 78 ms 108 ms 
average: 95
via ByteArrayInputStream: 
102 ms 100 ms 99 ms 75 ms 80 ms 76 ms 79 ms 82 ms 77 ms 76 ms 78 ms 78 ms 78 ms 76 ms 85 ms 76 ms 76 ms 79 ms 79 ms 77 ms 76 ms 102 ms 76 ms 77 ms 75 ms 75 ms 76 ms 75 ms 76 ms 79 ms 
average: 80
via Immutable: 
91 ms 69 ms 69 ms 68 ms 66 ms 64 ms 77 ms 67 ms 65 ms 62 ms 62 ms 63 ms 65 ms 83 ms 62 ms 63 ms 62 ms 66 ms 66 ms 73 ms 67 ms 67 ms 63 ms 63 ms 66 ms 68 ms 63 ms 65 ms 70 ms 64 ms 
average: 67

blacelle:
Same pattern to other containers does not improve performance on Issue319 gist . But I guess this is because Issue319 gist generate mainly BitmapContainers

lemire:
That's nice work. I guess that there is no catching up to "via Immutable" because we don't get the array directly, so there is extra monkeying around.

blacelle:
Given code is not ready: for RunContainers and ArrayContainers, I guessed the buffer for BitmapContainer was big enough. I do not know enough constrains on these containers to know if the constraint is relyable, or if a growing buffer should be considered.

There is also org.roaringbitmap.KryoTest.roaringTest() which fails, but I can not find why right now.

lemire:
No worry.

blacelle:
With RoaringBitmap.deserialize(ByteBuffer)

recommended: 
256 ms 82 ms 75 ms 143 ms 74 ms 73 ms 97 ms 74 ms 76 ms 99 ms 136 ms 80 ms 74 ms 76 ms 76 ms 137 ms 74 ms 74 ms 75 ms 75 ms 74 ms 129 ms 192 ms 74 ms 75 ms 77 ms 81 ms 83 ms 83 ms 108 ms 
average: 96
via ByteArrayInputStream: 
102 ms 101 ms 91 ms 76 ms 77 ms 77 ms 78 ms 79 ms 76 ms 75 ms 76 ms 80 ms 78 ms 75 ms 80 ms 76 ms 75 ms 75 ms 75 ms 76 ms 75 ms 84 ms 75 ms 74 ms 76 ms 76 ms 77 ms 75 ms 89 ms 75 ms 
average: 79
via Immutable: 
94 ms 67 ms 66 ms 65 ms 63 ms 67 ms 66 ms 62 ms 62 ms 65 ms 63 ms 64 ms 65 ms 71 ms 68 ms 67 ms 63 ms 66 ms 61 ms 63 ms 88 ms 65 ms 66 ms 65 ms 63 ms 66 ms 63 ms 64 ms 67 ms 71 ms 
average: 66
via Deserialize ByteBuffer: 
83 ms 71 ms 62 ms 59 ms 62 ms 93 ms 61 ms 60 ms 59 ms 60 ms 59 ms 60 ms 60 ms 73 ms 60 ms 60 ms 60 ms 60 ms 61 ms 60 ms 59 ms 65 ms 61 ms 61 ms 67 ms 62 ms 60 ms 59 ms 64 ms 62 ms 
average: 63

richardstartin:
It might be worth doing a comparison against #321

richardstartin:
I hadn't realised that this PR already had a `ByteBuffer` implementation virtually the same as in #321, and this one was first so I will close that PR. However, I have reservations about the changes made to the method which uses a `DataInput` here. They are made on the basis of observations about one implementation of `DataInput`, which don't generalise to other implementations see [the comment with benchmarks on the issue](https://github.com/RoaringBitmap/RoaringBitmap/issues/319#issuecomment-473516728) . They also require that potentially very large (tens of megabytes) `byte[]` arrays are allocated. I would like to see this PR merged, but would, as a user of the method being modified, first like to see:

* The changes to `serialize(DataInput)` reverted
* The new method `serialize(ByteBuffer)` kept and promoted as the most efficient option, and tested with a range of byte orders and buffer varieties
* A proper benchmark, using JMH, in the benchmarks project, **not** in the tests.

blacelle:
Perf Comparison with #321 

recommended: 
282 ms 89 ms 81 ms 193 ms 83 ms 82 ms 97 ms 75 ms 77 ms 109 ms 153 ms 89 ms 78 ms 81 ms 77 ms 78 ms 83 ms 75 ms 76 ms 75 ms 79 ms 123 ms 143 ms 93 ms 76 ms 79 ms 78 ms 78 ms 77 ms 78 ms 
average: 97
via ByteArrayInputStream: 
135 ms 99 ms 90 ms 75 ms 77 ms 77 ms 78 ms 84 ms 114 ms 76 ms 79 ms 79 ms 78 ms 78 ms 76 ms 78 ms 81 ms 78 ms 76 ms 76 ms 77 ms 77 ms 77 ms 78 ms 104 ms 78 ms 78 ms 76 ms 78 ms 77 ms 
average: 82
via Immutable: 
92 ms 68 ms 73 ms 64 ms 65 ms 65 ms 64 ms 65 ms 63 ms 65 ms 97 ms 65 ms 61 ms 63 ms 62 ms 63 ms 65 ms 65 ms 102 ms 66 ms 63 ms 64 ms 64 ms 65 ms 66 ms 66 ms 63 ms 66 ms 65 ms 66 ms 
average: 68
via Deserialize ByteBuffer: 
69 ms 71 ms 61 ms 58 ms 60 ms 67 ms 63 ms 62 ms 57 ms 60 ms 59 ms 59 ms 60 ms 58 ms 61 ms 59 ms 61 ms 59 ms 61 ms 62 ms 61 ms 61 ms 73 ms 62 ms 61 ms 63 ms 62 ms 63 ms 61 ms 59 ms 
average: 61
via Deserialize ByteBuffer (321): 
90 ms 73 ms 89 ms 65 ms 68 ms 66 ms 65 ms 68 ms 69 ms 77 ms 65 ms 66 ms 66 ms 65 ms 65 ms 66 ms 68 ms 67 ms 69 ms 66 ms 67 ms 72 ms 66 ms 66 ms 68 ms 65 ms 76 ms 68 ms 69 ms 66 ms 
average: 69

richardstartin:
The implementations are virtually identical so I would be surprised if there were any difference that couldn't be explained away as noise, and I'm not advocating for the use of the version in #321 because it's virtually the same thing. 

However, the way you are measuring performance is flawed - could you write a benchmark using JMH to make sure the benchmarks run in forked JVMs, warm up properly, measure over a range of inputs, and so on?

blacelle:
> They also require that potentially very large (tens of megabytes) byte[] arrays are allocated

I suggest adjusting the implementation to rely on a single byte[] based on BitmapContainer size. For the other containers, if the buffer is too small, it could be used to load by chunks (instead of 1 by 1 in original implementation, or all in a single byte[] in current proposition).

@richardstartin Yes, I feel the simpler would be to port the JMH benchmark from #320 . For now, the original gist was enough to pin the performance issue.

richardstartin:
Let me outline again why I would prefer that this PR does not modify the old method and simply introduces the new method:

- The benchmark evidence is based on `DataInputStream`, not `RandomAccessFile` etc. on a server class machine (i.e. not on a probably encrypted HDD). I think the benchmark evidence actually just highlights how bad `DataInputStream`/`ByteArrayInputStream` are.
- The modified method used to create very little temporary memory but now it creates a lot of it, even if you allocate just 8kB as you suggest. This may be an issue for users when picking up the new version. @ppiotrow has previously claimed to have 5 million bitmaps in his application - this would imply 40GB temporary memory to deserialise his bitmaps, assuming the change to use a 8kB buffer is made, and may require a buffering solution... 
- The new method which takes a `ByteBuffer` as input is just a much, much better solution, and supports something similar to the previous `RandomAccessFile` use case simply via `MappedByteBuffer`. In fact, it's exactly the buffering solution you would need to make point 2 work.

@blacelle I would prefer that you do not disregard my concerns as a user because an increase in memory consumption may cause me problems, but do note that I'm looking forward to having access to the new method.

@lemire please take my concerns as a user into account, and I suggest it may be worthwhile to collect feedback from other users too, given the space-time trade off being proposed here, before merging the modified method.

blacelle:
@richardstartin Nobody disregards your feedback. This PR has been open to showcase one way to address the original penalty. For now, I feel nothing is decided regarding `org.roaringbitmap.RoaringArray.deserialize(DataInput)`

blacelle:
About @ppiotrow use-case, I wonder if it would rely on .deserialize(ByteBuffer), or if the GC impact of the buffer would be counter-balancer by the performance issue revealed in #319

richardstartin:
The point I’m trying to make is that isn’t your call to make. This method is being used, and you want to make a space time trade off on the behalf of users. If this is merged, I will need to undertake unplanned work if I update for a bug fix. Why do you want to make this change? Because it affects your application? Or because a performance issue was raised and you want to fix it? If it’s the latter, I reiterate that introducing the new method and changing the documentation is the right course of action.

ppiotrow:
I've been called and my use cases are unaffected by additional memory allocation. We use `RoaringBitmap` (non-buffer version) only for offline building of bitmaps to serialise into file. In online scoring part we allocate large off heap buffer based on the file. Later we slice it and assign to `ImmutableRoaringBitmap` using this constructor https://github.com/RoaringBitmap/RoaringBitmap/blob/master/roaringbitmap/src/main/java/org/roaringbitmap/buffer/ImmutableRoaringBitmap.java#L967

richardstartin:
@blacelle it's not OK to copy and paste from my fork.

lemire:
@blacelle 


* I think that the introduction of `deserialize(ByteBuffer)` seems to please everyone. We probably just want to clean its comment by adding a few things. I think that the state of the `ByteBuffer` should be documented, after and before the call to `deserialize`. Does `limit` need to be set to the end of the serialized data, exactly? For the ImmutableRoaringBitmap construction, we make no such hard requirement... (it is just enough that there is enough data) Also, we should document whether the provided ByteBuffer is modified by the call. I would also add a code example, to show how one might go from a byte[] array to a bitmap via this method.

* There are concerns that the changes made to the existing `deserialize(DataInput)` method because of the allocation of an extra 8 kB of temporary data. Given that the existing `deserialize(DataInput)` method is a known quantity and that we have had no bug reported against it in quite some time... some level of conservatism is not insane. I can certainly create a scenario where the new `deserialize(DataInput)` method  will do poorly. Imagine that you have thousands of tiny (less than 8kB) bitmaps. It may well be that the cost of allocating the 8 kB buffers could start to cost you (mostly at the level of the garbage collector?).  What do you say about the following compromise... we keep the existing `deserialize(DataInput)` method and we introduce a `bufferedDeserialize(DataInput)` method (the name can be improved) where it is explained that we allocate a temporary 8kB buffer with each call, but that we expect better performance in some cases. The general idea is to 'push' the engineering up to the user. If someone does not care about the performance of `deserialize`, they can just go on with the existing one and be happy. If they would like to test out a better/faster approach, then they can adopt a new method. Evidently, in the existing `deserialize` method, we would refer to the other (possibly faster) method.
 

My proposal would increase the API size... but not a whole lot... and the performance benefits can be huge so it seems warranted.

I would also suggest that the PR should update our examples...

https://github.com/RoaringBitmap/RoaringBitmap/blob/master/examples/SerializeToByteArrayExample.java

blacelle:
I feel comfortable with keeping `serialize(DataInput)` as it stands, and provider an alternative taking advantage of an internal buffer (current code has impacted `.deserialize` and introduced a`.legacyDeserialize` to enable JMH benchmarks, but it is certainly safer to maintain `.deserialize` name to legacy behavior). I'm unsure if we should propose an optional input buffer to `.bufferedDeserialize`. Maybe going to `.deserialize(DataInput, byte[])` is the right way to propose this alternative (switching to 8kb buffer if input is null).

About the case with an input bitmap much smaller than 8kb, we can consider a more advanced policy : e.g. introduce a buffer on first BitmapContainer or RunContainer|ArrayContainer > 8kb, and rely on legacy behavior as long as we encounter only small RunContainer|ArrayContainer.

For symmetry, should we introduce a serialize(ByteBuffer) ?

blacelle:
About documentation, `org.roaringbitmap.buffer.ImmutableRoaringArray.ImmutableRoaringArray(ByteBuffer)` indicates the ByteBuffer is effectively copied by `java.nio.ByteBuffer.slice()`. However, I would say the ByteBuffer is not copied, but a new ByteBuffer is created, which shares its content with the input ByteBuffer. Then, the input ByteBuffer must not be mutated as long as the ImmutableRoaringArray is alive. And ImmutableRoaringArray will not mutate input ByteBuffer (neither in its position/mark/limit, nor in its content).

richardstartin:
@lemire 

Performance is a multidimensional matter. I could easily contrive a benchmark to demonstrate that uncompressed bitsets are "faster" than roaring for, say, XOR, on some datasets, and submit a PR. I hope that you would reject it, because the question is at what cost does the improvement come, and how general is the benefit? This PR imposes a memory consumption cost on unknown existing code, for what is sometimes a significant time improvement (but not on every data set, and over what timescale?). The correct course of action is not to make that tradeoff on the behalf of the user, but to add the new method. As a user I put trust in the maintainer to keep the library stable, and in return contribute bug fixes. So deprecate the old one, if you must, but please don't change things people depend on. I am convinced by your suggestion to add a new method that my interests as a user are being respected, so won't labour the point further.

Regarding copying and pasting, I hope not to appear to be petty, but there are widespread standards for attribution of work when code is written collaboratively on the Internet, and I am not the only person on GitHub who would object to code being taken without attribution via a merged commit.

lemire:
> I feel comfortable with keeping serialize(DataInput) as it stands, and provider an alternative taking advantage of an internal buffer. I'm unsure if we should propose an optional input buffer to .bufferedDeserialize. Maybe going to .deserialize(DataInput, byte[]) is the right way to propose this alternative (switching to 8kb buffer if input is null).

I am guessing that you'd like to see .bufferedDeserialize(InputStream) along with .bufferedDeserialize(InputStream, byte[]) where the user can provide their own buffer, if the provided buffer is to small at some point, then we just allocate a temporary buffer.

Is that it? 

> About the case with an input bitmap much smaller than 8kb, we can consider a more advanced policy : e.g. introduce a buffer on first BitmapContainer or RunContainer|ArrayContainer > 8kb, and rely on legacy behavior as long as we encounter only small RunContainer|ArrayContainer.

I am concerned that this will introduce code complexity and possibly some extra GC work for hard-to-measure benefits. The approach you describe above where you offer your own byte buffer seems better.

> For symmetry, should we introduce a serialize(ByteBuffer) ?

I have opened this as a separate issue: https://github.com/RoaringBitmap/RoaringBitmap/issues/322

lemire:
@richardstartin 

> Performance is a multidimensional matter. 

I agree.

> The correct course of action is not to make that tradeoff on the behalf of the user, but to add the new method. 

I agree.

> Regarding copying and pasting, I hope not to appear to be petty, but there are widespread standards for attribution of work when code is written collaboratively on the Internet, and I am not the only person on GitHub who would object to code being taken without attribution via a merged commit.

@blacelle : I suggest that you edit the code taken from Richard's PR so that the authorship of said code is clear to all. Presumably, using the 'author' tag from javadoc would do?

lemire:
> About documentation, (...)

We can certainly improve the existing documentation as far as ByteBuffer usage goes, but I would prefer to do it as separate issues. Want to open a new issue?

My concern here is purely instrumental with respect to the current PR: we want to tell the user what happens with the ByteBuffer they provide.

blacelle:
> I am guessing that you'd like to see .bufferedDeserialize(InputStream) along with .bufferedDeserialize(InputStream, byte[]) where the user can provide their own buffer, if the provided buffer is to small at some point, then we just allocate a temporary buffer.

For now, either there is no buffer and I allocate one if input Container > 8kb, else I rely on input buffer and I stand with it (leaving to the user the choice of a relevant buffer size).

> I am concerned that this will introduce code complexity and possibly some extra GC work for hard-to-measure benefits. The approach you describe above where you offer your own byte buffer seems better.

I suggest having a look to the coming commit (in a few minutes). It does not imply more GC and the additional complexity seems reasonable. The part I'm not fully confident yet is the loading by chunks, given a buffer smaller than the Container.

blacelle:
> We can certainly improve the existing documentation as far as ByteBuffer usage goes, but I would prefer to do it as separate issues. Want to open a new issue?

I simply added a precision than what will not mutate is the position/mark/limit/order of input ByteBuffer. I formulated this point as I copy-pasted the relevant documentation from ImmutableRoaringBitmap in this PR code.

lemire:
> I simply added a precision

Sorry... I think I misunderstood.

lemire:
> I suggest having a look to the coming commi

I think we all prefer to look at actual code. :-)

richardstartin:
@lemire 
> > The correct course of action is not to make that tradeoff on the behalf of the user, but to add the new method.
> 
> I agree.
> 

Thank you! Dropping out now.

lemire:
@blacelle Looks sensible to me. We seem to  have some dead code in RealDataSerializationBenchmark.java

lemire:
Let us see what travis-ci has to say... :-)

blacelle:
> Presumably, using the 'author' tag from javadoc would do?

@richardstartin Would it do?

blacelle:
Unclear why https://travis-ci.org/RoaringBitmap/RoaringBitmap/builds/508076463 is reported as failed

lemire:
Isn't programming fun? Thanks for your remarkable patience @blacelle.

blacelle:
@lemire Any hint why this appears in this PR?

blacelle:
In jdk8, no error. In jdk11, NPE in javadoc plugin...

lemire:
@blacelle Hard work!

Can you add some kind of attribution to Richard in jmh/src/main/java/org/roaringbitmap/realdata/RealDataSerializationBenchmark.java or wherever else it is needed in your opinion. I don't think we need to seek Richard's explicit approval... My belief is that he will be happy with a sensible acknowledgment.


> In jdk11, NPE in javadoc plugin...

Hmmm? Are you serious?

blacelle:
Screenshot 2019-03-18 at 23 47 00

blacelle:
Javadoc 3.1.0 mvn plugin works better...

lemire:
Ok. I take it that you recommend merging?

blacelle:
A last cleanup before OK for merging on my side: if travis is OK, I'm OK

roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java
```diff
@@ -264,12 +273,13 @@ protected void copyRange(int begin, int end, int newBegin) {
   }
 
   /**
-   * Deserialize.
+   * Deserialize. If the DataInput is available as a byte[] or a ByteBuffer, you could prefer
+   * relying on {@link #deserialize(ByteBuffer)}. If the InputStream is &gt;= 8kB, you could prefer
```
lemire: Is that &gt;= or &lt;=?

roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java
```diff
@@ -345,6 +355,287 @@ public void deserialize(DataInput in) throws IOException {
...
+   *
+   * @param in the DataInput stream
+   * @param buffer a byte[] used to buffer DataInput reading
+   * @throws IOException Signals that an I/O exception has occurred.
+   * @throws InvalidRoaringFormat if a Roaring Bitmap cookie
+   *             is missing.
+   */
+  public void deserialize(DataInput in, byte[] buffer) throws IOException {
+    if (buffer != null && buffer.length == 0) {
+      // Get ride of this useless buffer
```
lemire: I think that the correct spelling is "get rid of"?

roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
```diff
@@ -1429,6 +1430,24 @@ public boolean contains(long minimum, long supremum) {
...
 
 
+  /**
+   * Deserialize (retrieve) this bitmap.
+   * See format specification at https://github.com/RoaringBitmap/RoaringFormatSpec
+   *
+   * The current bitmap is overwritten.
+   *
+   * @param in the DataInput stream
+   * @param buffer a byte[] used to buffer DataInput reading
```
lemire: "The buffer gets overwritten with data during deserialization. You can pass a NULL value as a buffer. A buffer containing at least 8192 bytes might be ideal for performance. It is recommended to reuse the buffer between calls to deserialize (in a single-threaded context) for best performance."

lemire: That is, here I think we should provide guidance as to what "buffer" is and how it is supposed to be used.

roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java
blacelle: gt in &gt; stands for 'Greater Than'. Javadoc complained about '>'.

roaringbitmap/src/main/java/org/roaringbitmap/RoaringArray.java
blacelle: Fixed

roaringbitmap/src/main/java/org/roaringbitmap/RoaringBitmap.java
blacelle: Done

lemire:
End results is clean and passes all tests. Merging. I'll (try to) issue a release.

lemire:
Release Java software through maven has become crazily hard.

blacelle:
@lemire What are your latest issues regarding releases ?

lemire:
@blacelle I had to try three times.

1. You had updated the javadoc plugin to the latest version, this failed during release. I had to downgrade it. I am not sure why.

2. This solved the problem but then it just failed after uploading the package to maven. A time out arose. It wasn't on my end... my network connection was solid.

3. It worked.

Each time it creates a new, fake release. 

I would need to go in and rewrite git's history but I don't want to do this.

Note that this is part of why I reject all of the fancy tooling people want to sell me outside of the Java world. People want to sell me tools that automate C++ releases. No. F*cking way. 

Maven takes control of the process. It does not tell you what is wrong, and it just fails badly, without cleaning after itself.

It is really, really, really, really bad engineering. It is a disaster.

PR #411, opened 2020-06-20
a novel 64 bit roaring bitmap implementation

weijietong:
the novel 64 bit roaring bitmap implementation : 
* using [ART](https://db.in.tum.de/~leis/papers/ART.pdf) to keep the key/value pairs, key is the high 48 bit of the long value, value is the container id of the low 16 bit containers stored in `Containers`.
* serialization/deserialization of the art tree is native ,no need to reconstruct it.

@lemire please give a review


blacelle:
I ll try to get this soon through my use of 64b RoaringBitmap to check for bugs

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
```diff
@@ -0,0 +1,283 @@
...
+   * deserialize the node's body content
+   */
+  public abstract void deserializeNodeBody(DataInput dataInput) throws IOException;
+
+  public abstract int serializeNodeBodySizeInBytes();
+
+  /**
+   * find the position that the key and the node's prefix that begins to has a mismatch
+   */
+  public static int prefixMismatch(Node node, byte[] key, int depth) {
```
richardstartin, line 112: Let's use `Arrays.mismatch` via the Multi-Release shims jar.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
```diff
@@ -0,0 +1,283 @@
...
+        high = mid - 1;
+      } else {
+        return mid; // key found
+      }
+    }
+    // key not found.
+    return ILLEGAL_IDX;
+  }
+
+  private void serializeHeader(DataOutput dataOutput) throws IOException {
```
richardstartin, line 218: The preferred API for this since 0.8.0 serialises to a `ByteBuffer` because `DataOutput`/`DataInput` present lots of bottlenecks.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
```diff
@@ -0,0 +1,283 @@
...
+
+  private int serializeHeaderSizeInBytes() {
+    int size = 1 + 4 + 4;
+    if (prefixLength > 0) {
+      size = size + prefixLength;
+    }
+    return size;
+  }
+
+  private static Node deserializeHeader(DataInput dataInput) throws IOException {
```
richardstartin, line 237: ditto `ByteBuffer` preferable.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node16.java
```diff
@@ -0,0 +1,121 @@
+package org.roaringbitmap.art;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+
+public class Node16 extends Node {
+
+  byte[] key = new byte[16];
```
richardstartin, line 9: What about using 2 `long`s and saving the reference and the header/length?

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node16.java
```diff
@@ -0,0 +1,121 @@
...
+  Node[] children = new Node[16];
+
+  public Node16(int compressionLength) {
+    super(NodeType.NODE16, compressionLength);
+  }
+
+  @Override
+  public int getChildPos(byte k) {
+    return Node.binarySearch(this, k);
```
richardstartin, line 18: Could write a very efficient binary search on the bytes in 2 longs

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node256.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  public void replaceNode(int pos, Node freshOne) {
+    children[pos] = freshOne;
+  }
+
+  @Override
+  public int getMinPos() {
+    for (int i = 0; i < 256; i++) {
```
richardstartin, line 36: At the cost of a 256 bit bitmask, could we make this much cheaper?

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node256.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+    return ILLEGAL_IDX;
+  }
+
+  @Override
+  public int getNextLargerPos(int pos) {
+    if (pos == ILLEGAL_IDX) {
+      pos = -1;
+    }
+    for (pos++; pos < 256; pos++) {
```
richardstartin, line 49: A 256 bit bitmask would make this rank search more efficient with small memory overhead.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node256.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  @Override
+  public Node remove(int pos) {
+    this.children[pos] = null;
+    this.count--;
+    if (this.count <= 36) {
+      Node48 node48 = new Node48(this.prefixLength);
+      int j = 0;
+      for (int i = 0; i < 256; i++) {
```
richardstartin, line 94: again, 256 bitmask -> SWAR operation

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node256.java
```diff
@@ -0,0 +1,127 @@
...
+      if (children[pos] != null) {
+        return pos;
+      }
+    }
+    return ILLEGAL_IDX;
+  }
+
+  @Override
+  public int getMaxPos() {
+    for (int i = 255; i >= 0; i--) {
```
richardstartin, line 59: bitmask -> SWAR

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node4.java
```diff
@@ -0,0 +1,133 @@
+package org.roaringbitmap.art;
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class Node4 extends Node {
+
+  byte[] key = new byte[4];
```
richardstartin, line 10: Could use an `int` instead, save the header and the reference

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node4.java
```diff
@@ -0,0 +1,133 @@
...
+  Node[] children = new Node[4];
+
+  public Node4(int compressedPrefixSize) {
+    super(NodeType.NODE4, compressedPrefixSize);
+  }
+
+  @Override
+  public int getChildPos(byte k) {
+    for (int i = 0; i < count; i++) {
+      if (key[i] == k) {
```
richardstartin, line 20: cheaper on an `int`

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
```diff
@@ -0,0 +1,168 @@
...
+
+import java.io.DataInput;
+import java.io.DataOutput;
+import java.io.IOException;
+import java.util.Arrays;
+
+public class Node48 extends Node {
+
+  //the value of childIndex content won't be beyond 48
+  byte[] childIndex = new byte[256];
```
richardstartin, line 11: Since this is used for search operations where we skip over empty values, if this were `new long[32]` we could skip many bytes with a single instruction via `Long.numberOfTrailingZeros`

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
```diff
@@ -0,0 +1,166 @@
...
+  public HighLowContainer() {
+    art = new Art();
+    containers = new Containers();
+  }
+
+  public Container getContainer(long containerIdx) {
+    return containers.getContainer(containerIdx);
+  }
+
+  public ContainerWithIndex searchContainer(byte[] highPart) {
```
richardstartin, line 30: Can we use a `long` and an int `shift` instead? Reduce allocation.

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
```diff
@@ -0,0 +1,166 @@
...
+  @Override
+  public boolean equals(Object object) {
+    if (object instanceof HighLowContainer) {
+      HighLowContainer otherHighLowContainer = (HighLowContainer) object;
+      if (this.art.getKeySize() != otherHighLowContainer.art.getKeySize()) {
+        return false;
+      }
+      KeyIterator thisKeyIte = this.highKeyIterator();
+      while (thisKeyIte.hasNext()) {
+        byte[] thisHigh = thisKeyIte.next();
```
richardstartin, line 150: Could this return a `long` with the bytes in the appropriate positions instead?

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  @Override
+  public boolean equals(Object object) {
+    if (object instanceof HighLowContainer) {
+      HighLowContainer otherHighLowContainer = (HighLowContainer) object;
+      if (this.art.getKeySize() != otherHighLowContainer.art.getKeySize()) {
+        return false;
+      }
+      KeyIterator thisKeyIte = this.highKeyIterator();
```
richardstartin, line 148: It's not clear to me looking at this line of code alone whether this iterator is breadth or depth first, it was possible to vectorised equality checks by going breadth first in the 32 bit bitmap.

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
```diff
@@ -0,0 +1,82 @@
...
+    };
+    return high48;
+  }
+
+  //get the low 16 bit parts of the input data
+  public static char lowPart(long num) {
+    return (char) num;
+  }
+
+  public static long toLong(byte[] high, char low) {
```
richardstartin, line 23: I suggest extending the shims. On JDK9+ you can do this more efficiently with [`VarHandle`s](https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/MethodHandles.html#byteArrayViewVarHandle-java.lang.Class-java.nio.ByteOrder-) and for JDK8 users it's safe to use `Unsafe`

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  //compare according to the dictionary order
+  public static int compareHigh(byte[] a, byte[] b) {
+    return compareTo(a, 0, a.length, b, 0, b.length);
+  }
+
+  private static int compareTo(byte[] buffer1, int offset1, int length1,
+      byte[] buffer2, int offset2, int length2) {
+    if (buffer1 == buffer2 &&
```
richardstartin, line 66: This should use `Arrays.mismatch` via shims unless it's only intended for very short arrays

richardstartin:
Looks really good! I think there are a few places where we could avoid little `byte[]` being created, and get more operations per cycle and reduce spatial overhead in some of the node implementations (in `Node256` this would deviate from the ART paper but I expect it would work well to add a bitmask for child existence at very small cost). I think de/serialisation via `ByteBuffer` would be preferable.

weijietong:
thanks @richardstartin @blacelle. @richardstartin your advice is valuable. But converting node4, node16's byte[] to int or longs will make search ,sort, insert operations on the byte array be difficult.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
weijietong: agree ,will do that

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
weijietong: agree, I left that as TODO to make this PR come out soon. I will implement that at a later time.

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
weijietong: the high part byte array was fetched from the ART, no additional allocation

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
weijietong: no additional memory allocation ,not seen any benefit

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java
weijietong: the iterator is depth first model to visit all the keys from the minimum one to max one.

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
weijietong: agree

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
weijietong: This comparator different from the mismatch meaning. It's to return the compare result.

richardstartin:
What I had in mind really wasn’t very complicated at all actually. I can send you some changes later if you like.

weijietong:
@richardstartin I could try that but will be later as I am on vacation from now on. Never mind to show your example code. I wonder we will change the int or long into byte array some areas or through some shift operations.

weijietong:
@richardstartin  I spend some of my free time applying your review advice to the codes. Please give a review again, thanks!

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
weijietong: I am not familiar with this new API. I don't know how to use this API to accelerate the performance by assembling two different types array by looking at the docs. So I keep it.

weijietong:
need help! What's wrong with the `checkstyle` method name?
>
/home/travis/build/RoaringBitmap/RoaringBitmap/shims/src/main/java/org/roaringbitmap/ArraysShim.java:43:21: Method name 'mismatch​' must match pattern '^[a-z][a-z0-9][a-zA-Z0-9_]*$'. [MethodName]

lemire:
@weijietong  You have used a non-ASCII symbol in the method name. Please use only ASCII for names (this is what the style error message is telling you, effectively). I am surprised that Java will even compile this...

Screen Shot 2020-06-25 at 5 03 13 PM

weijietong:
@lemire thanks for your help,I have fixed that.

RoaringBitmap/src/main/java/org/roaringbitmap/Container.java
```diff
@@ -845,6 +852,13 @@ public Container or(Container x) {
    */
   public abstract void serialize(DataOutput out) throws IOException;
 
+  /**
+   * Serialize the container
+   * @param byteBuffer the ByteBuffer
+   * @throws IOException Signals that an I/O exception has occurred.
+   */
+  public abstract void serialize(ByteBuffer byteBuffer) throws IOException;
```
richardstartin, line 860: This duplicates `writeArray(ByteBuffer)`

RoaringBitmap/src/main/java/org/roaringbitmap/Container.java
```diff
@@ -263,6 +263,13 @@ public boolean contains(Container subset) {
    */
   public abstract void deserialize(DataInput in) throws IOException;
 
+  /**
+   * Deserialize the container
+   * @param byteBuffer the ByteBuffer
+   * @throws IOException Signals that an I/O exception has occurred.
+   */
+  public abstract void deserialize(ByteBuffer byteBuffer) throws IOException;
```
richardstartin, line 271: The reason this method didn't exist before was that the cardinalities and bitmap types are available in the serialised format of `RoaringArray` so they can be constructed there.

RoaringBitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
```diff
@@ -414,6 +414,16 @@ public void deserialize(DataInput in) throws IOException {
     }
   }
 
+  @Override
+  public void deserialize(ByteBuffer byteBuffer) throws IOException {
+    this.cardinality = 0;
```
richardstartin, line 419: This is inefficient compared to the way this is done elsewhere (converting to `LongBuffer` and calling `LongBuffer.get(long[])`) because the bounds checks get eliminated that way. It also shouldn't be necessary to calculate the cardinality, should just have it stored in the serialised format.

RoaringBitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
```diff
@@ -1265,6 +1275,14 @@ public void serialize(DataOutput out) throws IOException {
     }
   }
 
+  @Override
+  public void serialize(ByteBuffer byteBuffer) throws IOException {
```
richardstartin, line 1279: duplicates `writeArray`

RoaringBitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
```diff
@@ -1133,6 +1144,15 @@ public void serialize(DataOutput out) throws IOException {
     }
   }
 
+  @Override
+  public void serialize(ByteBuffer byteBuffer) throws IOException {
```
richardstartin, line 1148: duplicates `writeArray`

RoaringBitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
```diff
@@ -359,6 +359,17 @@ public void deserialize(DataInput in) throws IOException {
     }
   }
 
+  @Override
+  public void deserialize(ByteBuffer byteBuffer) throws IOException {
+    this.cardinality = 0xFFFF & byteBuffer.getChar();
```
richardstartin, line 364: Why does this have a cardinality but the `BitmapContainer` doesn't? I don't agree with these methods because they seem specific to the ART high/low container, but contributors would be tempted to use them with `RoaringArray`

RoaringBitmap/src/main/java/org/roaringbitmap/RunContainer.java
```diff
@@ -2110,6 +2121,11 @@ public void serialize(DataOutput out) throws IOException {
     writeArray(out);
   }
 
+  @Override
+  public void serialize(ByteBuffer byteBuffer) throws IOException {
```
richardstartin, line 2125: Exactly... duplicated `writeArray`

RoaringBitmap/src/main/java/org/roaringbitmap/art/Art.java
```diff
@@ -0,0 +1,390 @@
...
+
+  public void deserializeArt(ByteBuffer byteBuffer) throws IOException {
+    root = deserialize(byteBuffer);
+  }
+
+  public LeafNodeIterator leafNodeIterator(boolean reverse, Containers containers) {
+    return new LeafNodeIterator(this, reverse, containers);
+  }
+
+  private void serialize(Node node, DataOutput dataOutput) throws IOException {
```
richardstartin, line 283: This method doesn't check if nodes are empty. Can empty nodes be found in the ART?

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
```diff
@@ -0,0 +1,353 @@
...
+      }
+    }
+    // key not found.
+    return ILLEGAL_IDX;
+  }
+
+  private void serializeHeader(DataOutput dataOutput) throws IOException {
+    //first byte: node type
+    dataOutput.writeByte((byte) this.nodeType.ordinal());
+    //non null object count
```
richardstartin, line 235: Good, I'm glad these have appropriate sizes now. 

Can empty nodes be serialised though? I think they should probably be pruned from the serialised output if they can be in the in memory representation. If empty nodes are impossible, the count can be stored in a single byte with zero mapped to 256.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Art.java
weijietong: The empty was checked at the `HighLowContainer.serialize`.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java
weijietong: There's no empty node. The count number is in `byte` range. As I explained in the count member variable comment. It's  short type for coding benefit. As many places need to use the count, if it's byte type, you need to cast it to unsigned int before you can use the count directly.

RoaringBitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
weijietong: I am confused by the `writeArray`. I initially want to use the writeArray. But I found that the `writeArray` implementation logic is different from the corresponding `serialize(DataOutput )`, see `ArrayContainer` implementation for example. I think the writeArray's implementation should have the same logic with the serialize(DataOutput ) version.  These's no corresponding deserialization version to `writeArray`. That's why I implemented the serialize/deserialize (ByteBuffer)

richardstartin:
Thanks for making some of the suggested changes. I especially like that the sizes of some of the serialised attributes have been made smaller. I have a branch where I removed the need for small `byte[]` (meaning things like the longest common prefix matches and binary searches can be accelerated) which I could apply on top of this, so don't worry about those.

I think the `ByteBuffer` serialisation could do with some work; this PR introduces duplicated methods to the containers, look generic but are really specific to ART and can't be used with RoaringArray. The motivation to do this is that the container's cardinality is not stored in the leaf node; storing the container cardinalities in the ART leaf node would be more consistent with `RoaringArray` and would lead to some small efficiencies in deserialising containers.

Specifically:

* Store a container cardinality next to the container type when serialising leaf nodes.
* I would like to see the `Container.deserialize` methods moved into the ART package (make whichever `Container` constructors public you need to do this). 
* `Container.serialize` duplicates `writeArray` and is less efficient in `BitmapContainer` and `ArrayContainer` - remove these and call `writeArray` - once you have the cardinality in the leaf node you might as well do this anyway.

I appreciate it's probably frustrating to go back and forth making changes and meeting more requests for changes. I think the sooner we merge this the sooner it gets used and the sooner others can make improvements. So if you're OK to make the changes suggested, we can merge it to master afterwards, otherwise we could merge it to a development branch without the requested changes where others can make changes before making the feature generally available.

@lemire do you have any thoughts?

weijietong:
@richardstartin I am curious about how to keep compatible between the ArrayContainer's writeArray(ByteBuffer) and serialize(DataOutput) version. As their binary version data is not the same. When serializing the `Roaring64Bitmap` through the DataOutput , the ArrayContainer will have a cardinality data in the serialized data, If the user change to the `ByteBuffer` version to deserialize the data. It will be error.  Could we name the serialize(ByteBuffer) to a more specific name (maybe `serializeForArt`)?

richardstartin:
Sorry, I wasn't clear, both should be changed, because there's only one serialisation format :) the basic idea is just that the cardinality should be stored in the leaf node, for consistency, performance, and so on. Also, if this starts getting used seriously, a memory mapped version will need to be implemented on the same format, pushing the cardinality up to the leaf node would help with that. Does that make sense?

weijietong:
@richardstartin got it, thanks. Please review again.

lemire:
@weijietong 

Thanks for the good work (and thanks to @richardstartin for the thorough review).

I am happy that changes to the core classes is minimal.

I would like to invite you to improve slightly the test coverage before we merge...


There are several methods that are either untested or poorly tested...

- deserializeHeader in org/roaringbitmap/art/Node.java
- replaceChildren in org/roaringbitmap/art/Node256.java
- getCurrentContainerIdx() and replace(Container container) in org/roaringbitmap/art/ContainerIterator.java
- remove in org/roaringbitmap/longlong/HighLowContainer.java
- getNextSmallerPos, getMaxPos, replaceChildren, and all of the serialize* methods in org/roaringbitmap/art/Node16.java
- replace in org/roaringbitmap/art/Containers.java

Also the javadoc needs help...

```
> Task :shims:javadoc
RoaringBitmap/shims/src/main/java/org/roaringbitmap/ArraysShim.java:43: warning: no @param for a
  public static int mismatch(byte[] a, int aFromIndex, int aToIndex,
                    ^
RoaringBitmap/shims/src/main/java/org/roaringbitmap/ArraysShim.java:43: warning: no @param for b
  public static int mismatch(byte[] a, int aFromIndex, int aToIndex,
                    ^
RoaringBitmap/shims/src/main/java/org/roaringbitmap/ArraysShim.java:43: warning: no @return
  public static int mismatch(byte[] a, int aFromIndex, int aToIndex,
                    ^
3 warnings



> Task :RoaringBitmap:javadoc
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Shuttle.java:18: warning: no description for @return
   * @return
     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Art.java:92: warning: no @param for containers
  public KeyIterator iterator(Containers containers) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Art.java:92: warning: no @return
  public KeyIterator iterator(Containers containers) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Art.java:101: warning: no @param for key
  public long remove(byte[] key) {
              ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/ContainerIterator.java:96: warning: no @param for container
  public void replace(Container container) {
              ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:37: warning: no @param for k
  public abstract int getChildPos(byte k);
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:37: warning: no @return
  public abstract int getChildPos(byte k);
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:42: warning: no @param for pos
  public abstract Node getChild(int pos);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:42: warning: no @return
  public abstract Node getChild(int pos);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:47: warning: no @param for pos
  public abstract void replaceNode(int pos, Node freshOne);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:47: warning: no @param for freshOne
  public abstract void replaceNode(int pos, Node freshOne);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:52: warning: no @return
  public abstract int getMinPos();
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:59: warning: no @return
  public abstract int getNextLargerPos(int pos);
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:64: warning: no @return
  public abstract int getMaxPos();
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:69: warning: no @param for pos
  public abstract int getNextSmallerPos(int pos);
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:69: warning: no @return
  public abstract int getNextSmallerPos(int pos);
                      ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:76: warning: no @param for pos
  public abstract Node remove(int pos);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:147: warning: no @param for children
  public abstract void replaceChildren(Node[] children);
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:152: warning: no @param for dataOutput
  public abstract void serializeNodeBody(DataOutput dataOutput) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:152: warning: no @throws for java.io.IOException
  public abstract void serializeNodeBody(DataOutput dataOutput) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:157: warning: no @param for byteBuffer
  public abstract void serializeNodeBody(ByteBuffer byteBuffer) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:157: warning: no @throws for java.io.IOException
  public abstract void serializeNodeBody(ByteBuffer byteBuffer) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:162: warning: no @param for dataInput
  public abstract void deserializeNodeBody(DataInput dataInput) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:162: warning: no @throws for java.io.IOException
  public abstract void deserializeNodeBody(DataInput dataInput) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:167: warning: no @param for byteBuffer
  public abstract void deserializeNodeBody(ByteBuffer byteBuffer) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:167: warning: no @throws for java.io.IOException
  public abstract void deserializeNodeBody(ByteBuffer byteBuffer) throws IOException;
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:181: warning: no @param for childNode
  public static Node insertLeaf(Node current, LeafNode childNode, byte key) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:181: warning: no @param for key
  public static Node insertLeaf(Node current, LeafNode childNode, byte key) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:181: warning: no @return
  public static Node insertLeaf(Node current, LeafNode childNode, byte key) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:199: warning: no @param for src
  public static void copyPrefix(Node src, Node dst) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:199: warning: no @param for dst
  public static void copyPrefix(Node src, Node dst) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:210: warning: no @param for key
  public static int binarySearch(byte[] key, int fromIndex, int toIndex,
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:210: warning: no @param for k
  public static int binarySearch(byte[] key, int fromIndex, int toIndex,
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node.java:210: warning: no @return
  public static int binarySearch(byte[] key, int fromIndex, int toIndex,
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java:76: warning: no @return
  public LeafNodeIterator highKeyLeafNodeIterator(boolean reverse) {
                          ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/HighLowContainer.java:87: warning: no @param for highPart
  public void remove(byte[] highPart) {
              ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:8: warning: no @param for v
  public static byte[] toBDBytes(int v) {
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:8: warning: no @return
  public static byte[] toBDBytes(int v) {
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:20: warning: no @param for bytes
  public static int fromBDBytes(byte[] bytes) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:20: warning: no @return
  public static int fromBDBytes(byte[] bytes) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:30: warning: no @param for v
  public static int setByte(int v, byte bv, int pos) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:30: warning: no @param for bv
  public static int setByte(int v, byte bv, int pos) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:30: warning: no @param for pos
  public static int setByte(int v, byte bv, int pos) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:30: warning: no @return
  public static int setByte(int v, byte bv, int pos) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:39: warning: no @param for v
  public static int shiftLeftFromSpecifiedPosition(int v, int pos, int count) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:39: warning: no @param for pos
  public static int shiftLeftFromSpecifiedPosition(int v, int pos, int count) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:39: warning: no @param for count
  public static int shiftLeftFromSpecifiedPosition(int v, int pos, int count) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:39: warning: no @return
  public static int shiftLeftFromSpecifiedPosition(int v, int pos, int count) {
                    ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:48: warning: no @param for v
  public static byte firstByte(int v) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/IntegerUtil.java:48: warning: no @return
  public static byte firstByte(int v) {
                     ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java:56: warning: no @param for v
  public static byte[] toBDBytes(long v) {
                       ^
RoaringBitmap/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java:56: warning: no @return
  public static byte[] toBDBytes(long v) {
                       ^
52 warnings
```


Both of these issues (coverage and javadoc) should be fixed. 

The coverage issue may seem like nitpicking since you already added quite a few tests... but you are also adding quite a bit of code... We have had trouble many times with untested code that looked ok, went into production and caused no end of trouble... Of course, it is not because it is tested that it is correct, but it is much less embarrassing when there are failures and we can trace them back to tested code... it is then simply a matter of adding more tests. 

The javadoc fixes are just flat out necessary before we can do a release for obvious reasons. 

You refer on a few occasion to performance...  Would you consider contributing a benchmark... See  https://github.com/RoaringBitmap/RoaringBitmap/tree/master/jmh/src

Without any benchmarking, it is hard for people to appreciate the value of this new code. It would be really helpful for the users to see the performance...  It is not super easy to add a JMH benchmark and run it, but we have a ton of existing examples...

Cheers!

weijietong:
ok, please wait.

lemire:
@weijietong Note that it is ok to *remove* code so as to avoid coverage issues.

weijietong:
@lemire  I think it's ready to review, but there's two build errors which I can not get any meaningful message. Please give some help.

lemire:
I will check.

PR #464, opened 2021-03-02
Implement advanceIfNeeded for batch iterator

Cheappie:
Hello, I gave It a try to implement advanceIfNeeded for batch iterator. Do you see It beneficial to merge these changes ? Feedback is very welcome

### DIFF
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
@@ -51,9 +51,6 @@ public final class ArrayBatchIterator implements ContainerBatchIterator {
    * @param minVal - expected minimal value
    */
   public void advanceIfNeeded(char minVal) {
-    int newIndex = Util.advanceUntil(array.content, index - 1, array.getCardinality(), minVal);
-    if (index < newIndex) {
-      index = newIndex;
-    }
+    index = Util.advanceUntil(array.content, index - 1, array.getCardinality(), minVal);
   }
 }
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
@@ -1,7 +1,6 @@
 package org.roaringbitmap;
 
 
-
 public final class ArrayBatchIterator implements ContainerBatchIterator {
 
   private int index = 0;
@@ -29,7 +28,7 @@ public final class ArrayBatchIterator implements ContainerBatchIterator {
   @Override
   public ContainerBatchIterator clone() {
     try {
-      return (ContainerBatchIterator)super.clone();
+      return (ContainerBatchIterator) super.clone();
     } catch (CloneNotSupportedException e) {
       // won't happen
       throw new IllegalStateException(e);
@@ -44,5 +43,17 @@ public final class ArrayBatchIterator implements ContainerBatchIterator {
   void wrap(ArrayContainer array) {
     this.array = array;
     this.index = 0;
+  }
+
+  /**
+   * Advance iterator such that next value will be greater or equal to minVal
+   *
+   * @param minVal - expected minimal value
+   */
+  public void advanceIfNeeded(char minVal) {
+    int newIndex = Util.advanceUntil(array.content, index - 1, array.getCardinality(), minVal);
+    if (index < newIndex) {
+      index = newIndex;
+    }
   }
 }
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
@@ -14,9 +14,12 @@ public final class ArrayBatchIterator implements ContainerBatchIterator {
   public int next(int key, int[] buffer) {
     int consumed = 0;
     char[] data = array.content;
-    while (consumed < buffer.length && index < array.getCardinality()) {
+
+    int remainingBits = Math.min(array.getCardinality() - index, buffer.length);
+    for (int i = 0; i < remainingBits; i++) {
       buffer[consumed++] = key + (data[index++]);
     }
+
     return consumed;
   }
 
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/BitmapBatchIterator.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/BitmapBatchIterator.java
@@ -68,6 +68,6 @@ public final class BitmapBatchIterator implements ContainerBatchIterator {
       wordIndex = newWordIndex;
     }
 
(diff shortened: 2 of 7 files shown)


RoaringBitmap/src/main/java/org/roaringbitmap/BitmapBatchIterator.java
```diff
@@ -54,4 +54,22 @@ void wrap(BitmapContainer bitmap) {
...
+   */
+  public void advanceIfNeeded(char minVal) {
+    int newWordIndex = minVal >> 6;
+
+    if (wordIndex < newWordIndex) {
+      word = bitmap.bitmap[newWordIndex];
+      wordIndex = newWordIndex;
+    }
+
+    if ((64 * wordIndex) + numberOfTrailingZeros(word) < minVal) {
```
lemire, line 71: Can you explain why you add this check here? It is likely to translate into a significant performance penalty.

lemire:
@Cheappie  At a glance, this looks good. Can you check by two comments?

RoaringBitmap/src/main/java/org/roaringbitmap/BitmapBatchIterator.java
Cheappie: I agree, It is worrying me too. The reason is to prevent clearing bits from current word if someone after going forward, tried to advance backwards.

advanceIfNeeded(it, 84);
advanceIfNeeded(it, 30);

I have created unit tests that cover this case:
name: testAdvanceIfNeededToStalePositionDontClearCurrentBits, in ContainerBatchIteratorTest class

alternative approach that I came up is to create mask

int mask = ~(-((newWordIndex - wordIndex) >>> 31));
word &= 0xFFFFFFFFFFFFFFFFL << ((minVal & mask) & 0x3F);

If anyone has idea for more performant creation of such mask or achieving same outcome, please share It with me

RoaringBitmap/src/main/java/org/roaringbitmap/RoaringBatchIterator.java
```diff
@@ -58,16 +60,23 @@ private void nextIterator() {
...
+        nextIterator((BitmapContainer) container);
+      } else if (container instanceof RunContainer) {
+        nextIterator((RunContainer) container);
       }
-      key = highLowContainer.getKeyAtIndex(index) << 16;
+
+      key = containerPointer.key() << 16;
     } else {
       iterator = null;
     }
```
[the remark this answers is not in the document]
Cheappie: do you mean blank line? I can remove it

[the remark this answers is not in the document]
lemire: It is fine it looked wrong on my screen but I think it is ok.

RoaringBitmap/src/main/java/org/roaringbitmap/BitmapBatchIterator.java
lemire: Can we catch it right before?

```Java
if(wordIndex > newWordIndex) {
  return; // nothing todo
} else  if (wordIndex < newWordIndex) {
 ... same as currently
}
```

It would still add a branch, so it might slow things down a bit, but it would just be a compare and jump instead of the more expensive routine you are currently doing.

Cheappie: Yes we can, we could extend this approach further and invert check to act only when wordIndex is lesser or equal to newWordIndex. What do you think ?


    int newWordIndex = minVal >> 6;

    if (wordIndex <= newWordIndex) {
      if (wordIndex < newWordIndex) {
        word = bitmap.bitmap[newWordIndex];
        wordIndex = newWordIndex;
      }

      word &= 0xFFFFFFFFFFFFFFFFL << (minVal & 0x3F);
    }

lemire: That looks even better.

Cheappie:
> @Cheappie At a glance, this looks good. Can you check by two comments?

Hi @lemire, I have removed that blank line and I have updated advanceIfNeeded implementation for BitmapBatchIterator. Recently I have pushed few more unit tests. What do you think about the current state ?

lemire:
Changed the basis to 1.0-SNAPSHOT.

lemire (approved):
Marked for 1.0 release.

lemire:
The Java 8 tests failed, I am re-running.

Cheappie:
hey, for loop approach seems a bit faster than while on my laptop.

**Before**

Benchmark                                                   (bufferSize)  (keys)  Mode  Cnt   Score   Error  Units
ArrayBatchIteratorBenchmark.array_batch_iterator_benchmark           256    4096  avgt    5  10.467 ± 0.141  ms/op
ArrayBatchIteratorBenchmark.array_batch_iterator_benchmark           256   16384  avgt    5  56.890 ± 2.547  ms/op

**After**

Benchmark                                                   (bufferSize)  (keys)  Mode  Cnt   Score   Error  Units
ArrayBatchIteratorBenchmark.array_batch_iterator_benchmark           256    4096  avgt    5   8.827 ± 0.089  ms/op
ArrayBatchIteratorBenchmark.array_batch_iterator_benchmark           256   16384  avgt    5  48.247 ± 0.602  ms/op

RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
```diff
@@ -14,9 +14,12 @@ public ArrayBatchIterator(ArrayContainer array) {
   public int next(int key, int[] buffer) {
     int consumed = 0;
     char[] data = array.content;
-    while (consumed < buffer.length && index < array.getCardinality()) {
+
+    int remainingBits = Math.min(array.getCardinality() - index, buffer.length);
+    for (int i = 0; i < remainingBits; i++) {
       buffer[consumed++] = key + (data[index++]);
```
lemire, line 20: Yes. It makes sense that it would be a bit faster, good catch!

lemire:
@Cheappie I am going to merge this to 1.0-SNAPSHOT if you are happy.

Note that we currently can't release new versions, see https://github.com/RoaringBitmap/RoaringBitmap/pull/456

Cheappie:
@lemire I feel satisfied, and It should work. So I think that we can merge It. I was looking for some other possible optimizations around iterators, but I haven't found any with significant impact. Only extracting multiplication in BitmapBatchIterator gave ~5% speed up for bitmaps with cardinality 32768, but sometimes JIT fails to optimize code and It gets much worse so I have dropped this idea.

RoaringBitmap/src/main/java/org/roaringbitmap/ArrayBatchIterator.java
Cheappie: thank you!

lemire:
Merging to 1.0-SNAPSHOT

PR #476, opened 2021-04-14
Switch publishing to maven central

marshallpierce:


build.gradle.kts
```diff
@@ -24,27 +26,38 @@ subprojects {
...
+        }
+
+        named<JacocoReport>("jacocoTestReport") {
+            reports {
+                // used by coveralls
+                xml.isEnabled = true
+            }
+        }
+
+        withType<Javadoc> {
```
marshallpierce, line 55: drive-by squelching of the javadoc warning output since it makes other warnings harder to spot

build.gradle.kts
```diff
@@ -168,7 +173,16 @@ release {
...
 
-tasks.afterReleaseBuild {
-    // bintray is being sunsetted
-    //dependsOn(tasks.named("bintrayUpload"))
+nexusPublishing {
+    repositories {
+        sonatype {
+            // sonatypeUsername and sonatypePassword properties are used automatically
+            // id found via clicking the desired profile in the web ui and noting the url fragment
+            stagingProfileId.set("TODO")
```
marshallpierce, line 181: @lemire will need the ID from you for this

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
```
marshallpierce, line 3: 7.0 will work with Java 16

build.gradle.kts
lemire: I do have a user account on oss.sonatype.org with user ID `lemire`.  Is that what goes there?

lemire:
There are interesting error messages....

```

UnknownClass.JUnit Jupiter > UnknownClass.executionError FAILED
    org.junit.platform.commons.JUnitException at EngineExecutionOrchestrator.java:114
        Caused by: org.junit.platform.commons.JUnitException at HierarchicalTestEngine.java:54
            Caused by: java.util.concurrent.ExecutionException at ForkJoinTask.java:1006
                Caused by: java.lang.OutOfMemoryError at NativeConstructorAccessorImpl.java:-2
                    Caused by: java.lang.OutOfMemoryError at SpinedBuffer.java:750
```

@marshallpierce Did you touch the memory budget?

marshallpierce:
No, I don't think I did anything that should affect memory usage. :thinking:

build.gradle.kts
marshallpierce: That would be too easy. If you go to https://oss.sonatype.org/#stagingProfiles and select the profile you want, the url will change to `https://oss.sonatype.org/#stagingProfiles;ab8c5618978d18` (for my `org.mpierce` profile). The `ab8c5618978d18` is the part we need. This took a while for me to figure out...

marshallpierce:
Any gradle version > 6.3 caused that OOME -- perhaps behavior around junit / forking jvms / ??? changed. Probably should figure that out at some point, but no need to tackle that in this PR I think.

blacelle:
For the record, I tried switching to Central with https://github.com/RoaringBitmap/RoaringBitmap/pull/456, without success (including due do discomfort with Gradle and Kotlin). This PR shall be closed if this leads to a success. Still, there might be relevant pointers as I tried relying on references supposed to enable a human-less release.

marshallpierce:
Good to know. Hypothetically, making `release` depend on `closeAndReleaseSonatypeStagingRepository` would achieve that, but in practice it hasn't been reliable enough to automate that way. Perhaps with the newer repo it would be workable.

build.gradle.kts
lemire: So maybe 3a809bf4297e26

See screenshot.

Screen Shot 2021-04-15 at 9 22 45 AM

marshallpierce: That looks like the right thing. Can you give this branch a try with:

`./gradlew -Pversion=0.0 publishToSonatype`

That should create a staging repo and populate it with a fake version's worth of artifacts. You can then try "closing" the repo to see what errors pop up, if any. Either way, you can then "drop" the repo to discard it.

You'll need the following in your `~/.gradle/gradle.properties`:

```
sonatypeUsername = ...
sonatypePassword = ...

signing.keyId = ...
signing.password = ...
signing.secretKeyRingFile = ...
```

The [Gradle signing plugin](https://docs.gradle.org/current/userguide/signing_plugin.html) has some info on how to set up keys. I know almost nothing about GPG (other than that it is widely scorned by cryptographers), and I was able to muddle through it. Hopefully you already have some signing keys set up.

lemire:
@marshallpierce I am happy to merge this into our main branch. Even if it is somehow broken, it can hardly be worse than our current state where we can't release at all due to bintray being dead.

My current plan would be to merge and then try to do a patch release.

marshallpierce:
OK, it won't do any harm to merge in its current state.

build.gradle.kts
lemire: Let me try.

lemire:
@marshallpierce Ok, the following took 4 mins to run but it did work.

```
./gradlew -Pversion=0.0 publishToSonatype
```

lemire:
Merging.

PR #484, opened 2021-05-14
fix javadoc comment to be name the correct operation happening.

simeonpilgrim:
The comment says OR but the operation is XOR, so fix that.


weijietong:
Thanks for the fix! LGTM

PR #485, opened 2021-05-14
Add Flip ranger operation to Roaring64Bitmap

simeonpilgrim:
We were wanting a Flip range operation, so copied the existing "could be improved" operation used in the 32 bit implementation.

Added a few helper functions to work around change from int for the high word, to byte[].

Added unit tests to cover the cases I could think of while developing.

As a side note question if there is a way to get an enumerable range from the ART then the algorithm can be transposed to stepping over the existing containers in the range, and injected the gaps, and altering the existing nodes. This would remove the polling/search of every node in the range.

### DIFF
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
@@ -125,4 +125,43 @@ public class LongUtils {
   public static long initWithFirst4Byte(int v) {
     return ((long) v) << 32;
   }
+
+  /**
+   * shift the long left by the container size amount so we can loop across containers by +1 steps
+   * @param num long being treated as unsigned long
+   * @return value shifted out of value space into container high part
+   */
+  public static long leftShiftHighPart(long num) {
+    return num >>> 16;
+  }
+
+  /**
+   * shift the long by right the container size amount so we use the value after have done our steps
+   * @param num uint48 to be shift back into uint64
+   * @return value shifted out of container high part back into value space
+   */
+  public static long rightShiftHighPart(long num) {
+    return num << 16;
+  }
+
+  public static int maxLowBitAsInteger() {
+    return 0xFFFF;
+  }
+
+  /**
+   * set the high 48 bit parts of the input number into the given byte array
+   *
+   * @param num the long number
+   * @param high48 the byte array
+   * @return the high 48 bit
+   */
+  public static byte[] highPartInPlace(long num, byte[] high48) {
+    high48[0] = (byte) ((num >>> 56) & 0xff);
+    high48[1] = (byte) ((num >>> 48) & 0xff);
+    high48[2] = (byte) ((num >>> 40) & 0xff);
+    high48[3] = (byte) ((num >>> 32) & 0xff);
+    high48[4] = (byte) ((num >>> 24) & 0xff);
+    high48[5] = (byte) ((num >>> 16) & 0xff);
+    return high48;
+  }
 }
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
@@ -127,20 +127,20 @@ public class LongUtils {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
```diff
@@ -125,4 +125,43 @@ private static int compareTo(byte[] buffer1, int offset1, int length1,
   public static long initWithFirst4Byte(int v) {
     return ((long) v) << 32;
   }
+
+  /**
+   * shift the long left by the container size amount so we can loop across containers by +1 steps
+   * @param num long being treated as unsigned long
+   * @return value shifted out of value space into container high part
+   */
+  public static long leftShiftHighPart(long num) {
```
weijietong, line 134: Is the name right ? not `rightShiftHightPart` ?

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
```diff
@@ -125,4 +125,43 @@ private static int compareTo(byte[] buffer1, int offset1, int length1,
...
+    return num >>> 16;
+  }
+
+  /**
+   * shift the long by right the container size amount so we use the value after have done our steps
+   * @param num uint48 to be shift back into uint64
+   * @return value shifted out of container high part back into value space
+   */
+  public static long rightShiftHighPart(long num) {
```
weijietong, line 143: the same as above

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/Roaring64Bitmap.java
```diff
@@ -298,6 +298,61 @@ public void andNot(final Roaring64Bitmap x2) {
...
 
+  /**
+   * Complements the bits in the given range, from rangeStart (inclusive) rangeEnd (exclusive). The
+   * given bitmap is unchanged.
+   *
+   * @param rangeStart inclusive beginning of range, in [0, 0xffffffffffffffff]
+   * @param rangeEnd exclusive ending of range, in [0, 0xffffffffffffffff + 1]
+   */
+  public void flip(final long rangeStart, final long rangeEnd) {
+    //rangeSanityCheck(rangeStart, rangeEnd);
```
weijietong, line 309: please remove the comment code

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/Roaring64Bitmap.java
```diff
@@ -298,6 +298,61 @@ public void andNot(final Roaring64Bitmap x2) {
     }
   }
 
+  /**
+   * Complements the bits in the given range, from rangeStart (inclusive) rangeEnd (exclusive). The
+   * given bitmap is unchanged.
+   *
+   * @param rangeStart inclusive beginning of range, in [0, 0xffffffffffffffff]
```
weijietong, line 305: please add some tests to the range edge cases.

RoaringBitmap/src/test/java/org/roaringbitmap/longlong/TestRoaring64Bitmap.java
```diff
@@ -906,6 +906,158 @@ public void testAndNot_MultipleBucket() {
     assertEquals(123, left.select(0));
   }
 
+  @Test
+  public void testFlip_SameContainer() {
```
weijietong, line 910: please remove all the underline from the test method names

weijietong (commented):
thanks for your contribution! please address the comments.

RoaringBitmap/src/main/java/org/roaringbitmap/longlong/LongUtils.java
simeonpilgrim: OMG, you are so correct. Thank you for spotting..

RoaringBitmap/src/test/java/org/roaringbitmap/longlong/TestRoaring64Bitmap.java
simeonpilgrim: done

weijietong:
LGTM +1

PR #505, opened 2021-06-28
Fix Node48.getNextSmallerPos to handle input of 0 correctly

larsk-db:
### SUMMARY
- This adds a test case, which reproduces #504 
- Add a fix for the bug

### Automated Checks

### Discussion

It turns out this has nothing to do with the iterator seek code whatsoever, but it actually an ancient bug in `Node48.getNextSmallerPos`. It couldn't handle the case where `pos=0`. The `pos--;` before the loop would decrement it to `pos=-1`, which we would then right-shift to `i=536870911`. That's clearly not a legal index into a 48-value array.

Now this issue by itself an easy fix, but I have another concern about `Node48` implementation that I'd like some feedback on (I could fix both at the same time, ideally):
I feel like `Node48` is breaking an implicit contract that functions like `getChildPos`, `getNearestChildPos`, `getNextLargerPos`, `getNextLowerPos` should return a concrete position within `children`, not a key that still needs to be translated via `childrenIdx(pos, childIndex)` in `getChild`. All other `Node` implementations fulfil this contract (in `Node256` there's no difference between the two, of course).
I would like to make this implicit contract explicit, and convert `Node48` to adhere to it. I'm wondering if there is any particular reason that `Node48` was implemented differently? Is there any history to this, that I'm missing here? @weijietong @lemire @simeonpilgrim ?

### DIFF
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
@@ -145,7 +145,7 @@ public class Node48 extends Node {
     }
     pos--;
     int i = pos >>> INDEX_SHIFT;
-    for (; i >= 0 && i < 32; i--) {
+    for (; i >= 0 && i < LONGS_USED; i--) {
       long longv = childIndex[i];
       if (longv == INIT_LONG_VALUE) {
         //skip over empty bytes
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
@@ -143,6 +143,9 @@ public class Node48 extends Node {
     if (pos == ILLEGAL_IDX) {
       pos = 256;
     }
+    if (pos == 0) {
+      return ILLEGAL_IDX;
+    }
     pos--;
     int i = pos >>> INDEX_SHIFT;
     for (; i >= 0; i--) {
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
@@ -143,12 +143,9 @@ public class Node48 extends Node {
     if (pos == ILLEGAL_IDX) {
       pos = 256;
     }
-    if (pos == 0) {
-      return ILLEGAL_IDX;
-    }
     pos--;
     int i = pos >>> INDEX_SHIFT;
-    for (; i >= 0; i--) {
+    for (; i >= 0 && i < 32; i--) {
       long longv = childIndex[i];
       if (longv == INIT_LONG_VALUE) {
         //skip over empty bytes
--- a/RoaringBitmap/src/test/java/org/roaringbitmap/art/Node48Test.java
+++ b/RoaringBitmap/src/test/java/org/roaringbitmap/art/Node48Test.java
@@ -371,6 +371,15 @@ public class Node48Test {
   }
 
   @Test
+  public void testGetNextPosShouldNotThrowOnLegalInputs() {
+    Node node = new Node48(0);
+    for (int key = 0; key < 256; key++) {
+      Assertions.assertEquals(Node.ILLEGAL_IDX, node.getNextSmallerPos(key));
+      Assertions.assertEquals(Node.ILLEGAL_IDX, node.getNextLargerPos(key));
+    }
+  }
+
+  @Test
   public void testSetOneByte() {
     long[] longs = new long[Node48.LONGS_USED];
 


weijietong:
Since different node types have different storage implementations , so the `position` has different meanings there. I have no ` implicit contract` at the design to the `position`. The implicit contract is `getChildPos, getNearestChildPos, getNextLargerPos, getNextLowerPos` method should be compatible with `getChild` at one node.

larsk-db:
> Since different node types have different storage implementations , so the `position` has different meanings there. I have no ` implicit contract` at the design to the `position`. The implicit contract is `getChildPos, getNearestChildPos, getNextLargerPos, getNextLowerPos` method should be compatible with `getChild` at one node.

Yeah, that is kind of my point. I would like to upgrade the contract to what you said *plus*: "`getChild` is a cheap lookup operation". And also state it explicitly in the `Node` docs.

I think this will make the code 

a) easier to understand when a new person is reading it,  
b) faster (e.g., why do we do `childrenIdx` in both `getChildPos` and `getChild`? We should remember the result by returning it from `getChildPos`.)
c) less error-prone (e.g., who would expect `getChildKey` to just return you the value you gave it? Maybe someone wants to use it to know the key of the first child and does `getChildKey(0)`. This is going to be an awkward bug...).

weijietong:
@larsk-db  It would be good you do that.

weijietong:
I suggest you do that work at a separate PR.

larsk-db:
> @larsk-db It would be good you do that.

Ok, I'll get started on another PR for that. Shouldn't blow this one out of proportion.

> larsk-db Can you tackle a fix on top of this PR illustrating the bug?

Done, and also updated the title and description to reflect this. 
This makes this issue ready for review and then merge from my perspective.

lemire:
Let us wait for the tests to go green.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
```diff
@@ -143,6 +143,9 @@ public int getNextSmallerPos(int pos) {
     if (pos == ILLEGAL_IDX) {
       pos = 256;
     }
+    if (pos == 0) {
```
weijietong, line 146: pos is not illegal while it's 0. Keep compatible with `getChildKey`

larsk-db: Ah damnit. You see what I mean? I made exactly the mistake I literally just warned about. I'm also an idiot :D

larsk-db: But then again, what *should* it return? There is no smaller index to return. And returning `0` again doesn't seem right. It's not a `getNextSmallerOrEqualPos` method.

weijietong: suggest `if (pos > 0){pos--;}` or constraint the following `i` with a largest bound.

weijietong: that is `for (; i >=0 && i <32;i--)`

weijietong: You will also note that `getChildKey` will  benefit the performance from  current return `position` logic.

larsk-db: > that is `for (; i >=0 && i <32;i--)`

This has exactly the same effect as returning `ILLEGAL_INDEX` directly, since the loop will be skipped.



> suggest `if (pos > 0){pos--;}` 

I'm not convinced this is correct. Pretty sure is would implement `getNextSmallerOrEqualPos`.

weijietong:
I want to express that returning  ILLEGAL_INDEX  at that judgment would give a misunderstanding to others. I recalled the initial implementation it also constrained the i in the right range. So I suggest add checking logic at the for loop.

RoaringBitmap/src/main/java/org/roaringbitmap/art/Node48.java
```diff
@@ -145,7 +145,7 @@ public int getNextSmallerPos(int pos) {
     }
     pos--;
     int i = pos >>> INDEX_SHIFT;
-    for (; i >= 0; i--) {
+    for (; i >= 0 && i < 32; i--) {
```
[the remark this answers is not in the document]
weijietong: agree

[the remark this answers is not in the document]
larsk-db: Yes, of course. My bad.

simeonpilgrim (approved):
Looks good to me.

PR #527, opened 2021-10-06
Batch iterator skipping

richardstartin:


RoaringBitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
```diff
@@ -1771,20 +1771,8 @@ public void wrap(long[] b) {
...
-          return;
-        }
-        w = bitmap[x];
-      }
-    }
-    while (hasNext() && ((peekNext()) < (minval))) {
-      next(); // could be optimized
-    }
+    x = minval >>> 6;
+    w = bitmap[x] & -(1L << minval);
```
richardstartin, line 1775: On second thoughts, this is wrong, `next()` expects zeros to have been skipped over before calling it, but the tests pass, so this suggests a missing test case.

PR #634, opened 2023-07-12
Xor instead of subtraction

xtonik:
It is more performant to use XOR operation instead of subtraction using identity $`2^n - 1 - x = (2^n - 1) \oplus x`$ for $`x < 2^n`$.

### DIFF
--- a/RoaringBitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
+++ b/RoaringBitmap/src/main/java/org/roaringbitmap/BitmapContainer.java
@@ -1138,7 +1138,7 @@ public final class BitmapContainer extends Container implements Cloneable {
     }
     for (--x; x >= 0; --x) {
       if (bitmap[x] != 0) {
-        return x * 64 + 63 - Long.numberOfLeadingZeros(bitmap[x]);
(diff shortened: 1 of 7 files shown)


richardstartin:
I'd want to be sure the JIT compiler doesn't know this identity by taking a look at the disassembly before and after for at least one of these changes. Does this make a measurable difference?

lemire:
> I'd want to be sure the JIT compiler doesn't know this identity by taking a look at the disassembly before and after for at least one of these changes. 

The compiler needs to know that the x is smaller than 2^n, and then it needs to decide to do the optimization.

> Does this make a measurable difference?

Well. I would be quite surprised if this optimization made a measurable difference.

Recent processors can retire 4 additions per cycle.

I don't want to be prematurely negative, but we are unlikely to be performance bounded by addition operations.

richardstartin:
> The compiler needs to know that the x is smaller than 2^n, and then it needs to decide to do the optimization.

but it's easy enough to check

lemire:
LLVM does it...

Capture d’écran, le 2023-07-12 à 13 57 48

lemire:
I am skeptical that this is faster so I am not going to merge right away. I am concerned that it may make the code harder to read with no benefit.

PR #641, opened 2023-07-12
Unified lowest set bit

xtonik:
Use the same expression for lowest set bit extraction within whole project as there exists two equivalent expressions  - `word & -word` and `word & (word - 1)` - just for correctness only.

Note that, both has the same performance as I do not provide benchmark for this.


xtonik:
Wrong assumption, correct identity is: `word & -word` and `word - (word & (word - 1))`.

PR #661, opened 2023-10-29
Roaring bitmap to JDK bit set conversion

xtonik:
### SUMMARY
Added missing routine for conversion RoaringBitmap to BitSet from JDK. Done in very simple - iterating value by value, it is far from optimal possible solution.

### Automated Checks


RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
```diff
@@ -1,24 +1,43 @@
...
-  // word
+  public static final int BLOCK_LENGTH = BitmapContainer.MAX_CAPACITY / Long.SIZE;
+
+  /**
+   * Converts an immutable roaring bitmap to JDK bit set.
+   *
+   * @param bitmap original bitmap
+   * @return bit set equivalent to roaring bitmap
+   */
+  public static BitSet bitSetOf(ImmutableRoaringBitmap bitmap) {
```
lemire, line 26: If we implement this for `ImmutableRoaringBitmap`, we should implement it for `RoaringBitmap`.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
```diff
@@ -1,24 +1,43 @@
...
+    // TODO optimize by iterating by containers and
+    // for RunContainer use:  bitSet.set(from, to);
+    // for BitmapContainer use:
+    // MappeableBitmapContainer bc = (MappeableBitmapContainer)
+    //    bitmap.getContainerPointer().getContainer();
+    // BitSet bitSet2 = BitSet.valueOf(bc.toLongArray());
+    // bitSet2.and(bitSet2);
```
lemire, line 33: I am ok with this suboptimal implementation but what happens when `it.next()` returns a negative value?

RoaringBitmap/src/test/java/org/roaringbitmap/buffer/TestBitSetUtil.java
```diff
@@ -234,4 +235,10 @@ private static ByteBuffer toByteBuffer(BitSet bitset) {
     return ByteBuffer.wrap(bitset.toByteArray());
   }
 
+  @Test
+  public void bitSetOf() {
+    ImmutableRoaringBitmap bitmap = ImmutableRoaringBitmap.bitmapOf(1, 2, 4);
```
lemire, line 240: Please add a test with something like...
```
ImmutableRoaringBitmap bitmap = ImmutableRoaringBitmap.bitmapOf(1, 2, 4, -1);
```

xtonik: I did not realize this basic fact, see [my proposed solution](https://github.com/RoaringBitmap/RoaringBitmap/pull/661/commits/56e8e6a10c9cd98f556b41154e34e7a9a34a8aac) - method will possibly return two bit sets, the second one for "negative" values shifted by Integer.MIN_VALUE not to be negative and can be hold by bit set. For obvious reasons bit set holding positive values including zero is returned always even being empty.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
```diff
@@ -1,24 +1,55 @@
...
+      int value = it.next();
+      if (value >= 0) {
+        bitSetPositive.set(value);
+      } else {
+        bitSetNegative.set(value - Integer.MIN_VALUE);
+      }
+    }
+    if (bitSetNegative.isEmpty()) {
+      return new BitSet[]{bitSetPositive};
+    }
```
lemire, line 50: What if bitSetPositive is empty?

It seems simpler to just return both bitsets...

xtonik: This would be much better than throw an exception when if someone wants to convert "negative" (as RoaringBitmap treats them as unsigned int), then he must do some expensive work before - split bitmap and reset high bit for negative values.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
```diff
@@ -1,24 +1,43 @@
...
   // a block consists has a maximum of 1024 words, each representing 64 bits,
   // thus representing at maximum 65536 bits
-  public static final int BLOCK_LENGTH = BitmapContainer.MAX_CAPACITY / Long.SIZE; //
-  // 64-bit
-  // word
+  public static final int BLOCK_LENGTH = BitmapContainer.MAX_CAPACITY / Long.SIZE;
+
+  /**
+   * Converts an immutable roaring bitmap to JDK bit set.
+   *
```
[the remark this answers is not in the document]
xtonik: I have created [some benchmark](https://github.com/RoaringBitmap/RoaringBitmap/commit/da1dead0ce8bb4bc7222869ca5c37c6b477edd1f) to optimize conversion. There is trade off between too much code and performance. **What variant do you suggest to choose? Should I create separate PR for this or it can be included here?**

| Benchmark | inputData | Mode | Cnt | Score | Error | Units |
| --- | --- | --- | --- | --- | --- | --- |
| iterateThroughContainers | CONSECUTIVE | avgt | 5 | 52.782 | ±3.963 | ms/op |
| iterateThroughContainers | RANDOM | avgt | 5 | 66.945 | ±8.262 | ms/op |
| negativeValuesSeparately | CONSECUTIVE | avgt | 5 | 62.152 | ±1.104 | ms/op |
| negativeValuesSeparately | RANDOM | avgt | 5 | 85.505 | ±12.954 | ms/op |
| negativeValuesSeparatelyRegisterPreviousValue | CONSECUTIVE| avgt | 5 | 49.961 | ±8.429 | ms/op |
| negativeValuesSeparatelyRegisterPreviousValue| RANDOM | avgt |5 | 185.567 | ±5.139 | ms/op |
| preallocated | CONSECUTIVE | avgt | 5 | 67.221 | ±1.658 | ms/op |
| preallocated | RANDOM | avgt | 5 | 78.673 | ±5.566 | ms/op |
| original | CONSECUTIVE | avgt | 5 | 75.124 | ±27.983 | ms/op |
| original | RANDOM | avgt | 5 | 84.717 | ±3.251 | ms/op |

RoaringBitmap/src/test/java/org/roaringbitmap/buffer/TestBitSetUtil.java
xtonik: [Done](https://github.com/RoaringBitmap/RoaringBitmap/commit/372c0b551716c33884b373c42a3112690b2801ca#diff-3eb787aff967a9a86a3c2fa49ff908ee1e1caf516b2002f9e5e6f53a05adcc14).

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
xtonik: What about to use their common ancestor `ImmutableBitmapDataProvider`?

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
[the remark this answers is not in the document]
xtonik: The variant using streaming API inspired by #556 was added afterwards and it is magically the fastest for consecutive values.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
lemire: `ImmutableBitmapDataProvider` is an interface.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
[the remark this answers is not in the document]
lemire: > There is trade off between too much code and performance. What variant do you suggest to choose?

My concern here was that the library user should know what to expect. If we provide a convenience function that is possibly slow, we should warn them... otherwise it can be considered a performance bug.

Not everything needs to be super optimized with complex code, but the user should get some warning if the performance can be poor.

lemire (changes requested):
We could merge this, but I'd like you to add a comment (maybe just one sentence) in the function description, so that the users know that this do bit-by-bit construction. Further, we should warn them that the memory usage of the result could be significant.

Further, we would want to have the equivalent for ImmutableRoaringBitmap.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
xtonik: I mean use one of OOP principle "code against interface, not implementation", but I you suggest, I will duplicate the code for `RoaringBitmap`.

RoaringBitmap/src/main/java/org/roaringbitmap/BitSetUtil.java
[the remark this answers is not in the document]
xtonik: And what variant you suggest we should use? As I wrote earlier, I have created [benchmarks](https://github.com/RoaringBitmap/RoaringBitmap/commit/da1dead0ce8bb4bc7222869ca5c37c6b477edd1f) for this issue, where are already optimized implementation. Why I ask you what to choose? If you look on benchmark results, there is no solution dominating both for tested cases - for consecutive values is the best variant using _streamingApi_ and for random values _iterateThroughContainers_. After that I can merge chosen one into this PR and do some code clean up - at least comments as they are wrong if I remember right.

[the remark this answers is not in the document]
lemire: Ok. The issue is that your implementation is fine, but you do bit-by-bit conversion.
If the RoaringBitmap contains a bitset container, for example. Then you can just use [valueOf](https://docs.oracle.com/javase/8/docs/api/java/util/BitSet.html#valueOf-long:A-)(long[] longs) to map directly, at a highspeed, the content of the Roaring container to a JDK BitSet. You are obviously aware of this because you have it in a comment.

I don't think we even need benchmarks, it is going to be much faster... 

Now. It is fine *not* implementing this, but I would request that you then document the performance in the document of the function. (E.g., "this function may set the bits one by one and may be computationally intensive".) That's important because some users might decide to systematically convert their Roaring bitmaps to BitSet. We need to warn them about the performance implications.

The Go roaring implementation and the C Roaring implementations have equivalent functionality btw.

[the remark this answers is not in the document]
lemire: Here is the C implementation: https://github.com/RoaringBitmap/CRoaring/blob/daad140f013cbb4e8b515da264403843d8dc1755/src/roaring.c#L3306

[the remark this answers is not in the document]
lemire: Here is the Go implementation:
https://github.com/RoaringBitmap/roaring/blob/88e9207dc61ee814891a2a7cce4aa8abf1fbac1f/roaring.go#L186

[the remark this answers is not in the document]
lemire: I am *not* asking you to use these sorts of implementations. They require more testing and more effort. I am just asking you to send the users a signal regarding how it is currently implemented.

lemire:
@xtonik Let me repeat my request. If I were a user of this library, I would assume that a library-provided function that converts a roaring bitmap into a BitSet would try, at least in some cases, to be highly efficient. The implementation you propose is fine, but it will have quite low efficiency in some instances. We need to warn the users so that they cannot complain later if their performance is poor.

PR #682, opened 2023-11-06
Check instance reference first when comparing two array containers

xtonik:
### SUMMARY
Inspired by equivalent [code in C/roaring](https://github.com/RoaringBitmap/roaring/blob/c79926f24cdcf0dc5fa59f688c308f0045e3b182/arraycontainer.go#L196). The question is if it is really worthy as comparing the same object is not common case.

### Automated Checks


richardstartin:
When would this actually happen? When comparing a bitmap to itself? Then push the reference equality check up to the bitmap level and don't make this change. If a user has the same container in two different bitmaps then there is a very serious bug whereby an update to one bitmap modifies another bitmap.

xtonik:
I agree with you as I mentioned this in description. The condition is not valid for any existing test. So this could be bug in C/roaring.

PR #761, opened 2024-11-15
fix the bsi topk bug

Tan-JiaLiang:
### SUMMARY
According to the [Bit-sliced index arithmetic](https://dl.acm.org/doi/10.1145/376284.375669), the topk if n too many rows in F, turn off n bits from E in F.

### Automated Checks

### DIFF
--- a/bsi/src/main/java/org/roaringbitmap/bsi/buffer/BitSliceIndexBase.java
+++ b/bsi/src/main/java/org/roaringbitmap/bsi/buffer/BitSliceIndexBase.java
@@ -2,6 +2,7 @@ package org.roaringbitmap.bsi.buffer;
 
 import org.roaringbitmap.BatchIterator;
 import org.roaringbitmap.IntConsumer;
+import org.roaringbitmap.IntIterator;
 import org.roaringbitmap.bsi.BitmapSliceIndex;
 import org.roaringbitmap.bsi.BitmapSliceIndex.Operation;
 import org.roaringbitmap.bsi.Pair;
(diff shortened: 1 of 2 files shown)


bsi/src/main/java/org/roaringbitmap/bsi/buffer/BitSliceIndexBase.java
```diff
@@ -324,7 +325,14 @@ public MutableRoaringBitmap topK(ImmutableRoaringBitmap foundSet, int k) {
 
     MutableRoaringBitmap F = ImmutableRoaringBitmap.or(G, E);
     long n = F.getLongCardinality() - k;
-    F.remove(0L, (long) F.select((int) n));
+    if (n > 0) {
```
fanson, line 328: according to algorithm 4 of this paper [Slicing the Dimensionality: Top-k Query Processing for High-Dimensional Spaces](https://link.springer.com/chapter/10.1007/978-3-662-45714-6_2), another tiny modification maybe better.

`
    long gCardinality = G.getLongCardinality();
    if (gCardinality < k) {
      long eCardinality = E.getLongCardinality();
      long n = eCardinality + gCardinality - k;
      LongIterator iterator = E.getLongIterator();
      while (iterator.hasNext() && n > 0) {
        E.removeLong(iterator.next());
        n--;
      }
    } else {
      E = null;
    }
    return Roaring64NavigableMap.or(G, E);
`
this could avoid unnecessary OR operation when the count  matches the target 'k'. (OR will 'null' will return immediately)

fanson: sorry, forget about my last comment.
(E is ImmutableRoaringBitmap, could not call `remove`)

Tan-JiaLiang: > when the gCardinality matches the target k

I think that gCardinality is never equal to the target k. Because G only asigns if n < k.
WDYT? cc @lemire

fanson: when k=0, then gCadinality == k.
another question, why we allow k=0? should we return quickly or throws exception when k=0?

Tan-JiaLiang: you are right, I think we can return quickly.

fanson:
@lemire @blacelle i think this is OK now, what do you think? cc @Tan-JiaLiang

Tan-JiaLiang:
@fanson Very thanks for review, this PR is ready to be merged.

PR #763, opened 2024-12-22
RangeBitmap between with context

richardstartin:
### SUMMARY
Adds an overload `RangeBitmap.between(long min, long max, RoaringBitmap context)` to `RangeBitmap.between(long min, long max)` to accept a context bitmap to be intersected with, consistently with other overloads. This allows intersection with another bitmap to be pushed down into the range evaluation, allowing much of the `RangeBitmap` to be skipped over.

This also uncovered a bug in the existing `betweenCardinality(long min, long max, RoaringBitmap context)` which has been fixed.

### Automated Checks


lemire (approved):
LGTM

PR #777, opened 2025-07-25
issue-773-don't mutate in isEmpty

mkeskells:
### SUMMARY
make isEmpty simpler and non mutating
related to https://github.com/RoaringBitmap/RoaringBitmap/issues/773

### Automated Checks

There was one broken test broken before I chnaged, and its unaffected ([TestSerializationViaByteBuffer](https://github.com/RoaringBitmap/RoaringBitmap/compare/classes/org.roaringbitmap.buffer.TestSerializationViaByteBuffer.html). [[14] 8, LITTLE_ENDIAN, true](https://github.com/RoaringBitmap/RoaringBitmap/compare/classes/org.roaringbitmap.buffer.TestSerializationViaByteBuffer.html#testDeserializeFromMappedFile(int,%20ByteOrder,%20boolean)[14]). I am running on windows

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/longlong/Roaring64NavigableMap.java
@@ -1483,14 +1483,17 @@ public class Roaring64NavigableMap implements Externalizable, LongBitmapDataProv
     return size;
   }
 
-  @Override
   /**
    * Returns true if the bitmap is empty (i.e., has no set bits).
-   * In general, it is a a mutator method due to caching: this function modifies
-   * the internal state of the bitmap.
    */
+  @Override
   public boolean isEmpty() {
-    return getLongCardinality() == 0L;
+    for (BitmapDataProvider bitmap : highToBitmap.values()) {
+      if (!bitmap.isEmpty()) {
+        return false;
+      }
+    }
+    return true;
   }
 
   @Override


blacelle (approved):
Without fixing the root issue in #773, this feels good by:
1. Fixing the scenario provided in #773 (hence helping in real-world scenarios)
2. Skipping cardinality-evaluation in `.isEmpty` which looks not legitimate if not necessary. (i.e. the provided implementation is *probably* faster is all scenarios (`empty`, `!empty`)).

blacelle:
Thanks @mkeskells

blacelle:
I believe `highToBitmap.isEmpty()` is sufficient indeed.

mkeskells:
I think I did that originally but one of the tests generated a empty register that was still mapped

I may be incorrect, it was quite late

blacelle:
```
public long getLongCardinality() {
      [...]
      int indexOk = ensureCumulatives(highestHigh());

      // ensureCumulatives may have removed empty bitmaps
      if (highToBitmap.isEmpty()) {
        return 0L;
      }
```

-> `// ensureCumulatives may have removed empty bitmaps` suggests it is legitimate to have empty entries in `highToBitmap ` ; we try to prune them, but there is no such guarantee. At first sight, `ensureCumulatives` seems complex enough there might not be a trivial way to prune as pre-step in `.isEmpty`.

---

`.removeLong` removes from `highToBitmap` if the bitmap turns empty.
`.flip` does not remove from `highToBitmap` if the bitmap turns empty.

Hence, it explains how there may be empty bitmaps in `highToBitmap`.

blacelle:
The following feels safe:

```
  public boolean isEmpty() {
    if (allValid) {
      return highToBitmap.isEmpty();
    } else {
      for (BitmapDataProvider bitmap : highToBitmap.values()) {
        if (!bitmap.isEmpty()) {
          return false;
        }
      }
    }
    // all bitmaps are empty: this is empty
    return true;
  }
```

lemire:
Running tests.

blacelle:
Thanks @lemire . I edited the code snippet. I feel we lacks unit-tests around these cases (bitmap turning empty), just due to the fact we got covered but an unrelated test-case (which is still unclear to me how it get in such a case).

And the amount of moving parts around these optimizations makes it not trivial (`allValid`, `sortedCumulatedCardinality`, `sortedHighs`, `latestAddedHigh`).

Moreover, my latest proposition will not preven the issue in https://github.com/RoaringBitmap/RoaringBitmap/issues/773, as it proposes to remove empty bitmaps while calling .isEmpty (hence mutating a `TreeMap` in a `read` operation, which feels bad regarding concurrent reads scenarios).

blacelle:
Switching from `TreeMap` to a `ConcurrentSkipListMap` may be a low-effort/low-risk with low RAM penalty, to circumvent these concurrent read scenarios. (low RAM penalty as if the Map was to be the major RAM consumer, `Roaring64Bitmap` would probably be a better alternative (`Roaring64NavigableMap` consuming generally less when the number of `high` is relatively low)).

blacelle:
> @blacelle Do you object to merging @mkeskells's PR as is ?

The PR as-is looks fine.

Issue #776, opened 2025-07-25
non functional optimsations for Roaring64Bitmap


mkeskells:
**Is your feature request related to a problem? Please describe.**
Roaring64Bitmap is suboptimal WRT memory retained, and CPU time for common operation

**Describe the solution you'd like**
Data structures should be not storing duplicate data, or data in inefficient structures, or algorithims that can be improved

I have several ideas to improve the structure, but dont know if this is already in progress

here are a few simple ideas, from a simple profiling of insertion, and looking at the code
retained memory size

- dont create new empty arrays in `art.Node`
- seperate `art.Node` that cant have prefix from those that can
- dont store prefixLength as well and a array of the prefixes 
- dont store count for `art.Nodes` that have a defined count

Cpu used
- dont sort using insertion sort when adding to a sorted array (seems to be approx 25% of the insertion time
- there is some dead code for leaf nodes (I believe)
- optimise isEmpty to not count

Happy to work (with someone who knows the library better than I do) on this

some of the chnages are simpler than other. SOme may be contentious. Some may chnage the serialised form

PR #778, opened 2025-07-25
make isEmpty faster

mkeskells:
### SUMMARY
dont count and compare with 0
just see if any of the registers are non empty

### Automated Checks

There was one broken test broken before I changed, and its unaffected ([TestSerializationViaByteBuffer](https://github.com/RoaringBitmap/RoaringBitmap/compare/classes/org.roaringbitmap.buffer.TestSerializationViaByteBuffer.html). [[14] 8, LITTLE_ENDIAN, true](https://github.com/RoaringBitmap/RoaringBitmap/compare/classes/org.roaringbitmap.buffer.TestSerializationViaByteBuffer.html#testDeserializeFromMappedFile(int,%20ByteOrder,%20boolean)%5B14%5D). I am running on windows

Part of https://github.com/RoaringBitmap/RoaringBitmap/issues/776

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/longlong/Roaring64Bitmap.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/longlong/Roaring64Bitmap.java
@@ -841,7 +841,17 @@ public class Roaring64Bitmap implements Externalizable, LongBitmapDataProvider {
 
   @Override
   public boolean isEmpty() {
-    return getLongCardinality() == 0L;
+
+    if (highLowContainer.isEmpty()) {
+      return true;
+    }
+    Iterator<Container> containerIterator = highLowContainer.containerIterator();
+    while (containerIterator.hasNext()) {
+      Container container = containerIterator.next();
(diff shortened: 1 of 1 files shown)


mkeskells:
> This is good, but why not...
> 
> ```java
> public boolean isEmpty() {
>   return highLowContainer.isEmpty();
> }
> ```

I was not sure of the invariance

All of the tests pass with that so I have pushed the suggestion

blacelle:
All mutating methods seems indeed to check for emptyness. There might be obscure edge-cases when deserializing (e.g. from non-Java implementations which might not prune empty containers).

mkeskells:
> All mutating methods seems indeed to check for emptyness. There might be obscure edge-cases when deserializing (e.g. from non-Java implementations which might not prune empty containers).

I han not considered the case of deserialisation from another form, or  from a previous version that didnt prune

The code doesnt seems to expect this invariance (e.g. trim check for empty())
if would be simple to add a post deserialisation check

```

PR #803, opened 2025-08-31
Remove LighLowContainer, Containers. Make LeafNode refer to a COntainer, rather than indexing via another component 

mkeskells:
### SUMMARY
Simlify and streamline the access to a Container, having it held directtly by a LeafNode, rather than a reference to the container in another structure
This also removes the inherent unbounded memory leaks in HighLowContainer, and the retained memory (**EDIT** example in the comments)
It speed up access, by removing a layer if indirection on access to most operation (Foreach is probably the only operation that isnt faster, and may be slightly slower)

As serialisation was bound to this structure there was a rewrite of the serialisation & deserialisation structure and code
The format of the serialised form changes (its smaller)
The serialised form address a version number, and old formats wit fail with a message to indicate the format isnt correct. This also allows for future version aif format changes again

For the Serialisation and deserialisation The code was moved closer to the abstraction being handled, except for the cases of the Containers. I didnt want to chnage the scope of this PR to change the 32 bit structure and code
I think that some code duplication could be removed in a followup as the `DataInput`/`DataOutput` and `ByteBuffer` code are similar, and could be unified, howerev that would IMO be too much for the scop of this PR

**EDIT**
Benchmarks and raw results are inclded now as a comment

I have added test for the serialisation format
### Automated Checks

At this is a larger PR I have structure then commits the allow review of some of the enabling chnages prior to the bulk of the change. Each commit was initially tested against the unit tests in the project


lemire:
@mkeskells You need to sync with the main branch.

@blacelle @richardstartin and others: would you chime in ?

> I did try to discuss this as a design issue, but all I go was "you need benchmarks". No design discussion

I have a general bias against far reaching pull requests. My bias is not unusual nor is it controversial. Please read again what I wrote: *This is a massive PR. Difficult to review. Before doing anything of the sort, please make sure you have good benchmark results. Changing thousands of lines of code is not something that should be done lightly.*


Let us go back to how you justify it: **It speed up access, by removing a layer if indirection on access to most operation**
and then **This also removes the inherent unbounded memory leak**.


These are two objective and important statements. I am asking you to document and justify these claims.

mkeskells:
MArking as Draft as I will not have time on this PR - to sync up, tidy up or benchmark for at least this week

mkeskells:
The majority of the new code, other than tests is for the change to the serialisation format (I think. Haven't counted then though)

That change is required by this PR but could be implemented against the current main branch. Its not seperate entirely, but mostly

Thus would make the serialized form smaller, and less tied to the internal memory structure than the current serialised form. The change is largely orthogonal, but dependent

It would also make the serialised form deterministic based on the content, which it isn't currently, as the memory leaks in Containers are reflected in the serialised form, and dehydrated form

The numbers are easier to review as well (serialised size, time, garbage). Retained size should be unaffected

blacelle:
Changing the serialization format certainly adds friction to this PR. Do you confirm this PR does break binary compatibility ? Or does it change the binary output because of the optimizations, while keeping compatibility?

https://github.com/RoaringBitmap/RoaringBitmap/pull/598 refers to some matters around Roaring64 serialization format. (It is mostly focused on Map-based, but it refers to ART ones too).

mkeskells:
copied the serialisation format to https://github.com/RoaringBitmap/RoaringBitmap/pull/805
I havent checked it, but it passes tests. Probably some dead code there still

WIll check it in more details later, but sharing for the conversation, but its very draft

lemire:
> Every time that a LafNode is removed (e.g. empty) the Containers null the reference bu doesnt recliam or reuse that slot

I do not consider this a memory leak. 

If we were to consider that it is a memory leak, then you'd have to consider the standard HashMap in Java to be plagued by memory leaks. The HashMap implementation does not include any mechanism to reduce the capacity of the table when keys are removed. The resize() method is designed to either initialize the table or increase its size (typically doubling it). There is no code path that creates a smaller table when the number of entries decreases.

This does not mean that the current design cannot be improved, of course.


> In terms of the access, it can be seem that most operation that need to do any bitmap operation operation traverses the (...)

I get that your claim is that you can remove an indirection and thus improve the performance. However, if that is your claim, then you should be able to support it with benchmark results.

blacelle:
> Its also not an equivalent comparison. Repeated addition and removal for HashMap doesn't grow memory use (it caps it to the maximum size). Repeated additional and removal for this structure leak more memory. Serialisation and deserialization clean the memory for a HashMap, it doesn't for this structure.

I did not analyzed the potential memory-leak. @mkeskells Could you provide a unit-test demonstrating the issue, so we can agree current behavior is acceptable or not (independantly of any design analysis). (Similarly to benchmarks on the performance aspect, we're just expecting factual elements before moving forwards on such evolutions). Such a unit-test would also prevent future regressions (standard software practise).

PR #816, opened 2026-01-08
Avoid intersecting with containers that are already empty

xtonik:
### SUMMARY
Avoids intersecting with containers that are already empty and removes an unnecessary check for empty containers.  

This change is inspired by the implementation provided in #815.

### Automated Checks

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/FastAggregation.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/FastAggregation.java
@@ -373,6 +373,7 @@ public final class FastAggregation {
     }
 
     RoaringArray array = new RoaringArray(keys, new Container[numContainers], 0);
+    outer:
     for (int i = 0; i < numContainers; ++i) {
       Container[] slice = containers[i];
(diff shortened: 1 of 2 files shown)


xtonik:
Good remark. I have, of course, repeatedly checked all possible variants of intersecting two bitmaps and concluded, that cardinality is computed in most of cases. If it is not, that means cardinality is -1 and thus `isEmpty()` returns `false`, so the test like [`FastAggregation.java#L389`](https://github.com/RoaringBitmap/RoaringBitmap/blob/5c2c253a49e9f483ce63359f1b659092f7380222/roaringbitmap/src/main/java/org/roaringbitmap/FastAggregation.java#L389) does not negatively affect the computation, because if container needs to be "repaired" afterwards, the test fails.

lemire:
@richardstartin Does the answer by @xtonik seems satisfying ? (I have not looked at the code.)

richardstartin (approved):
I’ve changed my mind, this is probably positive and a good performance improvement, on balance.

PR #818, opened 2026-01-22
attempting to publish to maven central using nmcp

lemire:
This is an attempt (maybe the 10th attempt) at allowing maven central releases based on a proposal by @snazi

We currently have the following secrets:

- GPG_KEY_NAME  
- GPG_PASSPHRASE  
- GPG_PRIVATE_KEY  
- MAVEN_PASSWORD  
- MAVEN_USER

We have an existing CI action that publishes to github actions when a release is created: https://github.com/RoaringBitmap/RoaringBitmap/blob/master/.github/workflows/releases.yml
I am not touching it (much).

Instead we add a `Create Release` action which would be manually triggered. It would then (eventually) trigger our existing release CI, and also a new one that would (hopefully) publish to maven central.

For good measure, we also add a snapshot CI action.

Whether this is right, I do not know.

Fixes https://github.com/RoaringBitmap/RoaringBitmap/issues/749

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -179,9 +179,11 @@ release {
 nmcpAggregation {
   allowDuplicateProjectNames.set(true)
   centralPortal {
-    username.value(provider { System.getenv("ORG_GRADLE_PROJECT_sonatypeUsername") })
-    password.value(provider { System.getenv("ORG_GRADLE_PROJECT_sonatypePassword") })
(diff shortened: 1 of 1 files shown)


lemire:
I try merging this PR and testing it out and it failed even after I tried many things.

So I am declaring it a dead-end.

lemire:
@lhotari @kevinjqliu It cannot work because one workflow using GITHUB_TOKEN cannot trigger the other workflows. So it was dead on arrival.

PR #819, opened 2026-01-23
Replace consecutive char array filling with copying of a pre-allocated array

xtonik:
### SUMMARY
Filling a char array with consecutive values can be done either using a simple loop or by copying a pre-allocated char array.
The latter approach is up to 5 times more performant, especially for large value ranges. This affects the performance of wide range of routines dealing with consecutive values directly in `ArrayContainer` or `RunContainer`, and indirectly impacts some other routines that call them. The trade-off of this approach is permanently occupied memory due to the pre-allocated char array, which consumes an additional 128 kB of memory at most.

This feature can be enabled by calling `CharRangeFiller.allocate(char)` for partial range allocation, or, for allocating the entire character range, the more suitable `CharRangeFiller.allocateCompletely()`. To free the memory, `CharRangeFiller.deallocate()` could be used.

A bitmap analyzer recommendation was added, although detection of many runs in `RunContainersStats.hasManyLongRuns() `could be implemented in a more comprehensive way.

**Benchmark: Isolated char array filling**

```
CharRangeFillerBenchmark.arraycopy          12  thrpt   10  69133.466 ±  3123.095  ops/ms
CharRangeFillerBenchmark.arraycopy          16  thrpt   10  71981.190 ± 12174.344  ops/ms
CharRangeFillerBenchmark.arraycopy         100  thrpt   10  64071.132 ±  5643.923  ops/ms
CharRangeFillerBenchmark.arraycopy        4096  thrpt   10   5213.053 ±    51.235  ops/ms
CharRangeFillerBenchmark.arraycopy       16384  thrpt   10    863.610 ±     9.211  ops/ms
CharRangeFillerBenchmark.iteratively        12  thrpt   10  86908.744 ±   702.798  ops/ms
CharRangeFillerBenchmark.iteratively        16  thrpt   10  70800.389 ±  1686.227  ops/ms
CharRangeFillerBenchmark.iteratively       100  thrpt   10  22331.669 ±   271.267  ops/ms
CharRangeFillerBenchmark.iteratively      4096  thrpt   10    609.741 ±     5.151  ops/ms
CharRangeFillerBenchmark.iteratively     16384  thrpt   10    150.446 ±     2.023  ops/ms
```

**Benchmark: Char array filling used in ArrayContainer creation by range**

```
Benchmark                                                 (length)   Mode  Cnt      Score      Error   Units
CharRangeFillerUsedInArrayContainerBenchmark.arraycopy          10  thrpt   10  45331.204 ± 1050.502  ops/ms
CharRangeFillerUsedInArrayContainerBenchmark.arraycopy         100  thrpt   10  32752.445 ± 1205.521  ops/ms
CharRangeFillerUsedInArrayContainerBenchmark.arraycopy        4096  thrpt   10   1060.396 ±   20.323  ops/ms
CharRangeFillerUsedInArrayContainerBenchmark.iteratively        10  thrpt   10  39073.542 ± 1183.561  ops/ms
CharRangeFillerUsedInArrayContainerBenchmark.iteratively       100  thrpt   10  13584.141 ±  304.668  ops/ms
CharRangeFillerUsedInArrayContainerBenchmark.iteratively      4096  thrpt   10    309.166 ±    5.933  ops/ms
```
Based on these benchmarks, the minimal array length for switching from a simple loop to copying a pre-allocated char array was determined empirically to be **12**. It is hard to determine this constant as it is varying by the time, Java version and the context of the use.

Comparison benchmarks are included in this PR for potential future reference.

Within this work the code in `MappeableContainer.toEfficientContainer()` was deduplicated at the same time.

### Automated Checks

### DIFF
--- a/jmh/src/jmh/java/org/roaringbitmap/CharRangeFillerBenchmark.java
+++ b/jmh/src/jmh/java/org/roaringbitmap/CharRangeFillerBenchmark.java
@@ -1,0 +1,57 @@
+package org.roaringbitmap;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.concurrent.TimeUnit;
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 5, timeUnit = TimeUnit.MILLISECONDS, time = 200)
+@Measurement(iterations = 10, timeUnit = TimeUnit.MILLISECONDS, time = 200)
+@BenchmarkMode(Mode.Throughput)
+@Fork(
+    value = 1,
+    jvmArgsPrepend = {
+      "-XX:+UseG1GC",
+      "-XX:-TieredCompilation",
+      "-XX:+AlwaysPreTouch",
+      "-ms4G",
+      "-mx4G"
+    })
+public class CharRangeFillerBenchmark {
+
+  @Param({
+    // "11", // set CharRangeFiller.USE_ARRAYCOPY_MIN_SIZE < 12
+    "12",
+    "16",
+    "100",
+    "4096",
+    "16384",
+  })
+  public int length;
+
+  public char[] a;
+
+  private static final int DESTINATION_OFFSET = 100;
+
+  @Setup
+  public void prepare() {
+    CharRangeFiller.allocate(Character.MAX_VALUE);
+    a = new char[length + DESTINATION_OFFSET];
+  }
+
+  @Benchmark
+  public void arraycopy(Blackhole blackhole) {
+    CharRangeFiller.fill(a, DESTINATION_OFFSET, 0, length);
+    blackhole.consume(a);
+  }
+
+  @Benchmark
+  public void iteratively(Blackhole blackhole) {
+    for (int i = 0, j = DESTINATION_OFFSET; i < length; i++, j++) {
+      a[j] = (char) i;
+    }
+    blackhole.consume(a);
+  }
+}
--- a/jmh/src/jmh/java/org/roaringbitmap/CharRangeFillerUsedInArrayContainerBenchmark.java
+++ b/jmh/src/jmh/java/org/roaringbitmap/CharRangeFillerUsedInArrayContainerBenchmark.java
@@ -1,0 +1,57 @@
+package org.roaringbitmap;
+
+import org.openjdk.jmh.annotations.*;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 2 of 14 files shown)


xtonik:
Is there something what I could additionally provide to reviewers to look at this task? E.g. complete list of affected API methods / complete call graph with the "checks", that all of them are already properly tested, more benchmarks, ...

My limitation is that I don't have the means to test this across different hardware platforms. It must be done by somebody else.

In fact, it is actually a good thing that this issue has been sitting here for a while. Looking at it again, I realized that the same approach can be applied to conditional loops as well, although the performance boost won't be as significant (according to my preliminary benchmarks). I will add this to the upcoming implementation.

richardstartin:
Sorry I missed this. I neither want to approve nor block this PR. My gut feeling is that no user will ever call `CharRangeFiller.allocate` because it's a very obscure API. I also feel that ~128KiB would be too much to allocate eagerly. The update code is not thread-safe but I can't imagine race conditions actually arising if it's too obscure a mechanism to actually use. I would have taken the benchmark results on faith because this is a fairly straightforward workaround (array copies are fast, some data patterns are useful) for poor compiler optimisations for certain loops, but it's not clear how much a user is likely to notice a speed up in a high level operation like an intersection from using this API. Is this addressing a bottleneck?

xtonik:
> Sorry I missed this. I neither want to approve nor block this PR. My gut feeling is that no user will ever call `CharRangeFiller.allocate` because it's a very obscure API. I also feel that ~128KiB would be too much to allocate eagerly. The update code is not thread-safe but I can't imagine race conditions actually arising if it's too obscure a mechanism to actually use.

**Thread safety and obscure API could be solved by lazy holder pattern for pre-allocated array, but user will loose chance of clean up the memory, is this behavior ok?** Recommendation on the basis of bitmap's statistics could be removed then.

> I would have taken the benchmark results on faith because this is a fairly straightforward workaround (array copies are fast, some data patterns are useful) for poor compiler optimizations for certain loops, but it's not clear how much a user is likely to notice a speed up in a high level operation like an intersection from using this API. Is this addressing a bottleneck?

There is wide variety of API entry points, where it will be used, so it is very probable, that the array will be always allocated - `negateRange(), flip(), not(), add(int, int)`, conversions from run or bitmap to array containers and others. I can add some benchmarks special for this purpose, but I think that affected entry points are already covered by existing benchmarks. Benchmark for "Char array filling used in ArrayContainer creation by range" is not enough where novel approach is up to 3 times faster?

**What I could also add to PR is that the same pattern can be used in conditioned loops, eg. in BitmapContainer.iand(RunContainer).**

**What I have done poorly, that CharRangeFiller doesn't provide methods for doing task not to defensively check the thresholds as they should be correct in many cases of its internal use. I could add it to PR too.**

PR #831, opened 2026-06-24
Reuse ART shuttle stack entries instead of allocating per node

nishantmehta:
### Problem

`AbstractShuttle` (the ART tree cursor used by `Roaring64Bitmap.select`, `rankLong`'s leaf
iteration, and ordered iteration) keeps a fixed-depth `NodeEntry[MAX_DEPTH]` stack, but allocated
a **new `NodeEntry` on every push**. Traversing the tree therefore allocates a `NodeEntry` per node
visited — e.g. `select` of a middle element walks many leaves — even though at most `MAX_DEPTH`
entries are ever live at once.

### Change

Park a reusable `NodeEntry` in each stack slot (allocated the first time a given depth is reached)
and reset it to defaults on each push. The reset values match the previous `new NodeEntry()`
defaults exactly (`position = ILLEGAL_IDX`, `visited = false`, `startFromNextSiblingPosition =
false`, `leafNodeNextSiblingKey = 0`), so behaviour is unchanged.

### Result

`Roaring64Bitmap` with 200k entries, allocation per op (`ThreadMXBean.getThreadAllocatedBytes`):

| op | before | after | delta |
|---|---:|---:|---:|
| `select` | 590.9 | 206.1 | **−65%** |
| iterator (full) | 3920 | 2512 | **−36%** |
| `contains` | 24.0 | 7.2 | −70% |
| `rankLong` | 321.4 | 282.3 | −12% |

### Testing

Full `:roaringbitmap:test` passes, including the `org.roaringbitmap.longlong.*` (select / rank /
iteration) and `org.roaringbitmap.art.*` suites.

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/art/AbstractShuttle.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/art/AbstractShuttle.java
@@ -87,10 +87,8 @@ public abstract class AbstractShuttle implements Shuttle {
       if (nextPos != BranchNode.ILLEGAL_IDX) {
         stack[depth].position = nextPos;
         depth++;
-        // add a fresh entry on the top of the visiting stack
-        NodeEntry freshEntry = new NodeEntry();
-        freshEntry.node = currentBranchNode.getChild(nextPos);
-        stack[depth] = freshEntry;
+        // reuse the stack slot at this depth instead of allocating a fresh entry
+        useEntry(depth, currentBranchNode.getChild(nextPos));
       } else {
         // current internal node doesn't have anymore unvisited child,move to a top node
         depth--;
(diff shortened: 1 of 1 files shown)


roaringbitmap/src/main/java/org/roaringbitmap/art/AbstractShuttle.java
```diff
@@ -253,6 +243,24 @@ private void findNextSiblingKeyOfLeafNode() {
     }
   }
 
+  // Reuse the NodeEntry already parked in the stack slot at depth d (allocating one only
+  // the first time a given depth is reached) and reset it to the same defaults a fresh
+  // NodeEntry would have. This avoids allocating a NodeEntry on every push while the
+  // shuttle traverses the tree (select/rank/iteration).
+  private NodeEntry useEntry(int d, Node node) {
```
lemire, line 250: Why do you return NodeEntry?

lemire:
Looks fine, merging.

PR #835, opened 2026-07-07
`ArrayContainer.nextValue(char)` throws `ArrayIndexOutOfBoundsException` on an empty container

novoj:
### Summary
`ArrayContainer.nextValue` does `return fromValue == content[cardinality - 1] ? ... : -1;`. For `cardinality == 0` that's `content[-1]` → AIOOBE. `Util.advanceUntil(...,-1, 0, ...)` returns `0`, so the `index == cardinality` branch is taken. Siblings already handle empty (`BitmapContainer.nextValue` → `-1`; `ArrayContainer.previousValue` → empty-safe).

### Reproduction
```java
@Test
void nextValueOnEmptyArrayContainerMustReturnMinusOne() {
  ArrayContainer empty = new ArrayContainer();
  assertEquals(-1, empty.nextValue((char) 5));   // actual: AIOOBE Index -1
}
```

### Suggested fix (one early-return guard, non-hot path)
```diff
  public int nextValue(char fromValue) {
+   if (cardinality == 0) { return -1; }
    int index = Util.advanceUntil(content, -1, cardinality, fromValue);
```
Same guard applies to the buffer twin `MappeableArrayContainer.nextValue` (~L1470).

### Related prior art
#313 fixed the *non-empty* off-the-end case but left `cardinality == 0` unguarded — this is the residual edge case.

### Automated Checks

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ArrayContainer.java
@@ -1177,6 +1177,9 @@ public final class ArrayContainer extends Container implements Cloneable {
 
   @Override
   public int nextValue(char fromValue) {
+    if (cardinality == 0) {
+      return -1;
+    }
     int index = Util.advanceUntil(content, -1, cardinality, fromValue);
     if (index == cardinality) {
       return fromValue == content[cardinality - 1] ? (fromValue) : -1;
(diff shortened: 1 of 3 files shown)


lemire:
This is an invalid report.

The reporter did not explain how they arrive at this issue from our public API.

novoj:
You're right, and I apologize — this report doesn't stand up. The nextValue fault only manifests on an empty ArrayContainer, and a live RoaringBitmap never holds one (empty containers are pruned on removal and behind the !isEmpty() guards throughout the set operations), so there's no public-API path that reaches it. I found it by reading the method in isolation and reproduced it by constructing the container directly, which is exactly the state your invariants forbid — I should have verified reachability through the public API before filing.

For context, not as an excuse: this came out of an agent-assisted sweep of the sources while we were forking part of RoaringBitmap, and I filed the findings too eagerly without that reachability check. I'll add that gate before opening any further reports. Thanks for taking the time to look, and sorry for the noise.

Issue #110, opened 2016-05-16
Critical: overflow in RoaringIntIterator:advanceIfNeeded


jandam:
Version: 0.6.17

java.lang.ArrayIndexOutOfBoundsException: -2147483648
	at org.roaringbitmap.RoaringArray.getContainerAtIndex(RoaringArray.java:339)
	at org.roaringbitmap.RoaringBitmap$RoaringIntIterator.nextContainer(RoaringBitmap.java:88)
	at org.roaringbitmap.RoaringBitmap$RoaringIntIterator.advanceIfNeeded(RoaringBitmap.java:97)

this.pos can be greater than RoaringBitmap.this.highLowContainer.size() when last containers are empty.
  pos is incremented in while cycle but nextContainer do nothing.

Workaround:
  Use standard RoaringIntIterator::hasNext/next

Suggested fix:
```
   public void advanceIfNeeded(int minval) {
            while((('\uffff' & this.hs >>> 16) < ('\uffff' & minval >>> 16)) && hasNext()) { // FIX for this issue
                ++this.pos;
                this.nextContainer();
            }

            if(this.hasNext()) {
                this.iter.advanceIfNeeded(Util.lowbits(minval));
                if(!this.iter.hasNext()) {
                    ++this.pos;
                    this.nextContainer();
                }
            }

        }

```

lemire:
Verified. New release is in production.

lemire:
Hopefully fixed in 0.6.18. Reopen if you disagree. Please consider adding a failing test if your choose to re-open.

jandam:
Thank you very much for fast release. Fix is working for me.

PR #836, opened 2026-07-07
`ReverseIntIteratorFlyweight` uses a `short` container index — reverse iteration silently yields nothing past 32768 containers

novoj:
### Summary
`ReverseIntIteratorFlyweight` stores its container index in `private short pos;` and narrows it in `wrap()`: `this.pos = (short)(size() - 1);`. A bitmap can hold up to 65536 containers; once the count exceeds `Short.MAX_VALUE + 1` (32768), the cast overflows to **negative**, `hasNext()` (`pos >= 0`) returns false immediately, and reverse iteration silently produces nothing. The forward `IntIteratorFlyweight` correctly uses `int pos` with no cast.

### Reproduction
```java
@Test
void reverseIterationYieldsEveryValueWithMoreThanShortMaxContainers() {
  int containerCount = Short.MAX_VALUE + 2;          // 32769
  int[] data = new int[containerCount];
  for (int i = 0; i < containerCount; i++) data[i] = i << 16;   // one value per container
  RoaringBitmap bitmap = RoaringBitmap.bitmapOf(data);
  ReverseIntIteratorFlyweight it = new ReverseIntIteratorFlyweight(bitmap);
  assertTrue(it.hasNext());                          // actual: false (silent empty)
  int seen = 0;
  while (it.hasNext()) { it.next(); seen++; }
  assertEquals(containerCount, seen);                // actual: 0
}
```

### Suggested fix (zero runtime cost — widens one field to match the forward counterpart)
```diff
-  private short pos;                                           // ReverseIntIteratorFlyweight.java
+  private int pos;
-  this.pos = (short) (this.roaringBitmap.highLowContainer.size() - 1);
+  this.pos = this.roaringBitmap.highLowContainer.size() - 1;
```
Same change in the buffer twin `BufferReverseIntIteratorFlyweight` (field ~L33, `wrap` ~L109).

### Related prior art
#110 — an overflow in the *forward* `RoaringIntIterator.advanceIfNeeded`; same family of bug, different class. No prior report exists for the reverse flyweight.

### Automated Checks

### DIFF
--- a/roaringbitmap/src/main/java/org/roaringbitmap/ReverseIntIteratorFlyweight.java
+++ b/roaringbitmap/src/main/java/org/roaringbitmap/ReverseIntIteratorFlyweight.java
@@ -24,7 +24,7 @@ public class ReverseIntIteratorFlyweight implements IntIterator {
 
   private ReverseRunContainerCharIterator runIter = new ReverseRunContainerCharIterator();
 
-  private short pos;
+  private int pos;
 
   private RoaringBitmap roaringBitmap = null;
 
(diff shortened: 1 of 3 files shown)


lemire:
Merging.

